PR #7, opened 2014-07-23
phantomjs support & remove obsolete code

dmacvicar:
Implements support for running the testsuite using phantomjs, which is a WebKit based browser created specially for testing and has some advantages like being completely headless.

Also, remove support for other browsers that are obsolete tools and the code is not even working and remove unnecessary complexity, leaving only Firefox and PhantomJS.

@mseidl :+1: ?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,5 +3,6 @@ source 'https://rubygems.org'
 gem "cucumber","~> 1.3.0"
 gem "capybara","~> 2.1.0"
 gem "selenium-webdriver"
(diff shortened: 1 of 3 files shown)


renner:
@mseidl, @dmacvicar: Can we close this meanwhile?

isbm:
I am wondering if this PR will naturally get decomposed into a sand and then due to the gravity, would spill out from this list on @mseidl 's table. :wink:

dmacvicar:
I see this commits are already in mater

PR #14, opened 2015-02-16
Let Cuke quit on first failure

isbm:
Sometimes you need to just stop when cuke got first error, because the next errors often might make no sense as being based on the success of the scenario that just failed.

Usage:
`export TERMINATE=yes`


isbm:
Merged

PR #15, opened 2015-02-16
Various testsuite HEAD fixes

dmacvicar:
Replace the "test that there are 3 admins links somewhere in the page" with
the specific user page assertion that one admin link is in the table.

For that, we add table step definition to test for links in a more natural
way.


isbm:
Merged

PR #18, opened 2015-07-29
Salt minion smoketests

isbm:
Add basic Salt Minion smoke test and activate it at the beginning.
Note: this test also cleaning up itself (i.e. un-registers the client key and restarts the minion) for the future tests.

### DIFF
--- a/features/salt.feature
+++ b/features/salt.feature
@@ -15,3 +15,25 @@ Feature: Check if SaltStack Master is configured and running
   Scenario: Check SaltStack Master is properly configured
     When I issue command "ss -nta | grep 9080"
     Then it should contain "127.0.0.1:9080" text
+
+  Scenario: Check SaltStack Minion is running
+    When I issue local command "rcsalt-minion status | grep Active"
+    Then it should contain "active (running) since" text
+
+  Scenario: Check SaltStack Minion can be registered
+    When I issue command "salt-key --list unaccepted"
+    Then it should contain testsuite hostname
+    When I issue command "yes | salt-key -A"
+    And when I issue command "salt-key --list accepted"
+    Then it should contain testsuite hostname
+
+  Scenario: Check if SaltStack Minion communicates with the Master
+    When I ping client machine from the Master
+    Then it should contain testsuite hostname
+    When I get OS information of the client machine from the Master
+    Then it should contain "SLES" text
+
+  Scenario: Cleaning up for the general testsuite
+    When I delete key of the testsuite hostname
+    And when I issue command "salt-key --list unaccepted"
+    Then it should contain testsuite hostname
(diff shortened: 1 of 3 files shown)


features/salt.feature
```diff
@@ -15,3 +15,25 @@ Feature: Check if SaltStack Master is configured and running
   Scenario: Check SaltStack Master is properly configured
     When I issue command "ss -nta | grep 9080"
     Then it should contain "127.0.0.1:9080" text
```
dmacvicar: I am not sure how a mortal would relate that "ss -nta | grep 9080" to the feature.

Especially if I go to the [Salt Documentation](http://docs.saltstack.com/en/latest/topics/tutorials/firewall.html) and I read that salt master listens on ports 4505 and 4506. 9080, by my own reasoning is the port you chose for the API. So this scenario is a) confusing b) incomplete.

I suggest this reads like documentation:

```cucumber
Scenario: Check SaltStack Master is properly configured
   Then the Salt rest-api should be listening on local port 9080
   And the salt-master should be listening on public port 4505
  And the salt-master should be listening on public port 4506     
```
Then in the *details* (the steps) you can translate the public/9080 to a cal to ss looking for  127.0.0.1:9080 and the call to public 4506 to ss greping for *:4506.

It is not different to the problem we had in other scenarios that leak the DOM stuff to the Scenario descriptions.

features/step_definitions/salt_steps.rb
```diff
@@ -1,9 +1,36 @@
...
 
 When(/^I get a content of a file "(.*?)"$/) do |filename|
   $output = sshcmd("cat #{filename}")
 end
 
+When(/^I issue local command "(.*?)"$/) do |command|
+  $local_output = `#{command}`
+  $output = {:stdout => $local_output}
+end
+
```
dmacvicar: Those two steps are not specific to salt and therefore they belong in a different file. If you follow the previous suggestions, then the step for "And the salt-master should be listening on public port XX" may belong here, and this step could be implemented with the step above, but the step above is not Salt specific, and I bet there is a similar step somewhere else.

features/step_definitions/salt_steps.rb
```diff
@@ -1,9 +1,36 @@
...
+When(/^I delete key of the testsuite hostname$/) do
```
dmacvicar: If you are referring to the client here, then use either what we use on other scenarios (I think we use on "this client") or in this case you can use "this minion".

features/step_definitions/salt_steps.rb
```diff
@@ -1,9 +1,36 @@
...
+  # Longest key re-issue is 10 seconds. We sleep here 15.
+  sleep(15)
+end
+
+When(/^I ping client machine from the Master$/) do
+  $output = sshcmd("salt #{$testsuite_hostname} test.ping")
+end
+
+When(/^I get OS information of the client machine from the Master$/) do
+  $output = sshcmd("salt #{$testsuite_hostname} grains.get os")
```
dmacvicar: I am not sure here, but I think the convention when passing data from one step to the next is to use @ member variables and not globals.

PR #22, opened 2015-08-27
Update tests for Minions page

isbm:
Tests for https://github.com/SUSE/spacewalk/pull/317
Should be merged after 317 is in the trunk.

### DIFF
--- a/features/step_definitions/minions_steps.rb
+++ b/features/step_definitions/minions_steps.rb
@@ -1,0 +1,20 @@
+Then(/^I should see a this client as a minion in the Pending section$/) do
+  fail if not find('#pending-list').find("b", :text => @this_client_hostname).visible?
+end
+
+When(/^when I accept minion$/) do
+  link = "/rhn/manager/minions/accept/#{@this_client_hostname}"
+  fail if not page.has_xpath?("//a[@href='#{link}']")
+  find('#pending-list').find(:xpath, "//a[@href='#{link}']").click
+  sleep(5)
+  visit "/rhn/YourRhn.do"
+  visit "/rhn/manager/minions"
+end
+
+Then(/^I should see a this client as a minion in the Accepted section$/) do
+  fail if not page.find('#accepted-list').find("b", :text => @this_client_hostname).visible?
+end
+
+When(/^when I see the contents of the minion$/) do
+  visit "/rhn/manager/minions/#{@this_client_hostname}"
+end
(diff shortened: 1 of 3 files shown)


features/step_definitions/minions_steps.rb
```diff
@@ -0,0 +1,57 @@
+def click_n_refresh(link)
+  fail if not page.has_xpath?("//a[@href='#{link}']")
+  find('#pending-list').find(:xpath, "//a[@href='#{link}']").click
+  sleep(5)
+  visit "/rhn/YourRhn.do"
+  visit "/rhn/manager/minions"
+end
+
+Then(/^I should see a this client as a minion in the Pending section$/) do
+  fail if not find('#pending-list').find("b", :text => @this_client_hostname).visible?
```
dmacvicar: I would not use fail here but assert, that we started using recently:

```ruby
assert find('#pending-list').find("b", :text => @this_client_hostname).visible?
```

isbm: Whilst we had this discussion, I am not convinced how this makes any significant changes to a boolean _visible_ or _invisible_. But I indeed would reconsider this for `assert`, after Cynthia hammers all the UI and this whole test will be changed. Let this stay for now like that.

dmacvicar: Simple: we decided, that moving forward, we will use assertions, so there is no point in introducing more fails.

Now, if you want details why it is better: feedback.

fail is a RuntimeError. It makes hard to see at first glance whether the test failed because there was an error while running it (eg. disk space?) or if the test really failed because the conditions where not met.

With fail we get:

```ruby
RuntimeError: 
```

(plus a backtrace). Was there a timeout? Did the harddisk ran out of space?

For a simple boolean assertion, in the worst case you will get this:

```ruby
assert(find('#pending-list').find("b", :text => @this_client_hostname).visible?)
```

```ruby
Test::Unit::AssertionFailedError: <false> expected but was <true>
```

It is already telling you that there was no "error", but the condition *WE* are testing did not pass. It already gives you more information.

Now, just write it using a higher-level assert:

```ruby
assert_predicate(find('#pending-list').find("b", :text => @this_client_hostname), :'visible?')
```

And you will get something like:

```ruby
Test::Unit::AssertionFailedError: <#<Node:0x00560be49d85f0>>.visible? is true value expected but was <false>
```

And there you have something like assert_equal messages, with minimal effort.

If we change that the next time we touch this tests, I am fine, but we should already understand why this is preferable.

PR #23, opened 2015-09-01
Check for a minion in the dashboard

isbm:
Merge after https://github.com/SUSE/spacewalk/pull/321


isbm:
@renner maybe. :wink:

PR #24, opened 2015-09-17
Tests for the visible packages

isbm:
Originally made for the minions, but suitable also for a old RHN systems.
It also adds forgotten "Minions Overview" test.


features/system_packages.feature
```diff
@@ -0,0 +1,31 @@
...
+    And I select "Management" from "baseentitlement"
+    And I press "Update Properties"
+
+  Scenario: Check installed packages are visible
+    When I follow "Systems"
+    Then I follow this client as a registered system
+    Then I follow "Software"
+    Then I follow "List / Remove"
+    Then I should see a "aaa_base" text
+    And I should see a "aaa_base-extras" text
```
renner: Is this new feature actually tested anywhere? I am missing a reference in `testsuite.yml` here and it should come after the feature that actually accepts the key. Is there actually such a feature yet that accepts the key?

features/minion_overview.feature
```diff
@@ -0,0 +1,16 @@
...
+
+Feature: Explore the Minions page
+  In Order to validate that the Minions page works
+  As an authorized user
+  I want to see all Minions
+
+  Background:
+    Given I am authorized as "admin" with password "admin"
+
+  Scenario: Minion is available in the Overview dashbord
```
renner: Typo: It should be "dashboard" I guess.

features/system_packages.feature
isbm: Yes. :wink:

features/step_definitions/system_packages.rb
```diff
@@ -0,0 +1,11 @@
...
+  step %[I should see a "#{@this_client_hostname}" text]
+end
+
+Then(/^I follow this client as a registered system$/) do
+  step %[I follow "#{@this_client_hostname}"]
+end
+
+Then(/^I press "(.*?)"$/) do |btn_title|
+  click_button(btn_title)
+end
```
renner: @isbm: All these step definitions already exist elsewhere so this part of the patch can effectively be dropped it seems (see [here] (https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/step_definitions/register_client_steps.rb#L53), [here] (https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/step_definitions/register_client_steps.rb#L64) and [here] (https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/step_definitions/navigation_steps.rb#L50)). Please make generally sure to not duplicate these with every new feature!

isbm: Actually, for some reasons the "I click ..." didn't worked.

renner:
I am worried a bit about how the new minion/salt integration related tests are integrated with the old tests for the "traditional" features. Remember that accepting a key actually registers the minion as a system, while there is also other tests verifying the traditional registration via bootstrap script. Will the traditional registration overwrite the registered minion later on? When the packages are being verified, are we looking at the packages reported from salt or the packages reported via the RHN stack?

As we want to keep all the old tests, I guess it might be better to keep the salt stuff separated to avoid this kind of confusion. Instead of merging the new tests throughout `testsuite.yml`, wouldn't it be better to run the salt related tests in a block at the end after all the traditional tests are done and any previously registered system entries have been removed? What do you think, @mcalmer, @mseidl, @dmacvicar?

isbm:
@renner it is actually a good point, IMO. Currently I think this test is "dirty", because client is running both libraries at the same time. IIRC, only few things are going to be clashed and in fact, this whole excercise should not differ very much at the end from the traditional RHN and Salt way from the UI perspective.

So the real problem is that we have currently _only one_ client, which is at the same time "traditional RHN" and "Salt Minion", while we're selling the idea that client can be _either_ one. 

I would propose an idea to rather bring another passive client, an SSH integrated child client from the Cucumber Testsuite main client. Since RHN is still actively around and Salt is only at the beginning of development, I would suggest to have Salt Minion as a passive client. That means in the QA setup we would have:

1. SUMA server (sumas-<version>.suse.de)
2. RHN client, running Cucumber testsuite (sumac-<version>-rhn.suse.de)
3. Minion client, just plain JeOS (sumac-<version>-mn.suse.de)

RHN-activated Cucumber test suite would run as usual. After common channels were sync'ed and all long lasting tasks for common use were done, RHN-activated machine can call things on Minion client and test that part in the Web UI.

Pros:

- No clashes between the systems
- Cleanly tested a machine that does not have RHN libraries at all
- Cleanly tested a machine that does not have SaltStack libraries installed at all
- More than one client tested :wink: 

Cons:

- More complex setup.

If there are other, better ways to achieve such separation?

mcalmer:
I like the idea of a seperate client.
If we use a plain JeOS image we could also test the bootstrapping method. (however this will be implemented)

renner:
I like the suggestion of a separate client as well, thanks for creating the [Trello card] (https://trello.com/c/709OPgZ5)! Please make sure though that in the meantime the testsuite is not broken because of that. At least it seems as if grouping together all the tests related to the salt integration would make sense anyway, and this is something that we can do already now (with this PR).

isbm:
Nothing is clashing at the moment.

renner:
It was actually clashing, see for instance [these test results] (http://milkyway.cloud.suse.de/pub/2015-09-22_19-01-SM-head-cucumber-results_x86_64.html), hopefully fixed with fcc8903.

I would actually suggest that we merge this patch now. Even if the test for the packages would still fail (until https://github.com/SUSE/spacewalk/pull/330 will be merged) this should fix the other ~100 tests.

isbm:
OK, let it fail until the merge.

PR #26, opened 2015-12-01
Test for the download endpoint

dmacvicar:
Requires: https://github.com/SUSE/spacewalk/pull/401 because it uses simple signed tokens using the server secret directly without any extra key derivation.

### DIFF
--- a/features/download_endpoint.feature
+++ b/features/download_endpoint.feature
@@ -1,0 +1,14 @@
+Feature: Endpoint to download packages
+  In Order distribute software to the clients
+  As an authorized user
+  I want to download packages from the channels
+
+  Scenario: user without token
+    Given I try download "virgo-dummy-2.0-1.1.noarch.rpm" from channel "sles11-sp3-updates-x86_64-channel"
+    Then the download should get a 403 response
+
+  Scenario: user with a valid token for the org
+    Given I have a valid token for organization "1"
+    Then I try download "virgo-dummy-2.0-1.1.noarch.rpm" from channel "sles11-sp3-updates-x86_64-channel"
+    Then the download should get no error
+
--- a/features/step_definitions/download_endpoint_steps.rb
+++ b/features/step_definitions/download_endpoint_steps.rb
@@ -1,0 +1,28 @@
+require 'open-uri'
+require 'tempfile'
+
+Given(/^I try download "([^"]*)" from channel "([^"]*)"$/) do |rpm, channel|
+  url = "#{Capybara.app_host}/rhn/manager/download/#{channel}/getPackage/#{rpm}"
+  if @token
+    url = "#{url}?#{@token}"
+  end
+  puts url
+  Tempfile.open(rpm) do |tmpfile|
+    @download_path = tmpfile.path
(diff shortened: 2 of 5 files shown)


features/download_endpoint.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Endpoint to download packages
+  In Order distribute software to the clients
+  As an authorized user
+  I want to download packages from the channels
+
+  Scenario: user without token
+    Given I try download "virgo-dummy-2.0-1.1.noarch.rpm" from channel "sles11-sp3-updates-x86_64-channel"
+    Then the download should get a 403 response
+
+  Scenario: user with a valid token for the org
```
lucidd: May be nice if we also test with an invalid token.

features/step_definitions/download_endpoint_steps.rb
```diff
@@ -0,0 +1,28 @@
+require 'open-uri'
+require 'tempfile'
+
+Given(/^I try download "([^"]*)" from channel "([^"]*)"$/) do |rpm, channel|
+  url = "#{Capybara.app_host}/rhn/manager/download/#{channel}/getPackage/#{rpm}"
```
isbm: Maybe just `$TESTHOST`?

dmacvicar: We set Capybara.app_host from TESTHOST. I prefer to use what is assigned, as it *may* have been sanitized, while TESTHOST is user input.

features/download_endpoint.feature
dmacvicar: Yes, I will do that as a next step.

lucidd:
otherwise this looks good to me :+1:

features/step_definitions/download_endpoint_steps.rb
```diff
@@ -0,0 +1,28 @@
...
+  if @token
+    url = "#{url}?#{@token}"
+  end
+  puts url
+  Tempfile.open(rpm) do |tmpfile|
+    @download_path = tmpfile.path
+    begin
+      open(url) do |urlfile|
+        tmpfile.write(urlfile.read)
```
isbm: I am not sure, but shouldn't you call `tmpfile.unlink` after its [finalization](http://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html)?

dmacvicar: >When a Tempfile object is garbage collected, or when the Ruby interpreter exits, its
>associated temporary file is automatically deleted. This means that’s it’s unnecessary to >explicitly delete a Tempfile after use, though it’s good practice to do so: not explicitly deleting
>unused Tempfiles
>can potentially leave behind large amounts of tempfiles on the filesystem until they’re
>garbage collected. The existence of these temp files can make it harder to determine a new
>Tempfile filename.

So, it is removed when garbage collected. It is a good practice to do it explicitly.

Finalizer is setup like this: https://github.com/ruby/ruby/blob/trunk/lib/tempfile.rb#L135

PR #28, opened 2016-03-17
Test that spacewalk-repo-sync is working

dmacvicar:
With bnc#970951 this is what the test would show:

screenshot from 2016-03-17 09-30-29

With the proposed PR:

screenshot from 2016-03-17 09-29-55


mseidl:
LGTM

PR #29, opened 2016-03-17
test for mgr-create-bootstrap-repo

dmacvicar:
Should fail in current HEAD

screenshot from 2016-03-17 09-59-49


mseidl:
Looks good

PR #30, opened 2016-03-23
test for disabled local repositories after registration (bnc971788)

dmacvicar:
This test checks that after minion registration, all local repositories are disabled (repos not starting with susemanager:)

https://bugzilla.suse.com/show_bug.cgi?id=971788


dmacvicar:
I merged this manually.

PR #38, opened 2016-05-10
New Package for node, fix hardcoded hostname

MalloZup:


features/support/env.rb
```diff
@@ -74,8 +74,7 @@ def debrand_string(str)
 end
 host = "https://#{host}"
```
kkaempf: Umm, converting a hostname to a url and calling it `host`. Violates the principle of least surprise.

features/support/env.rb
```diff
@@ -74,8 +74,7 @@ def debrand_string(str)
...
 
-$myhostname = `hostname -f`
-$myhostname.chomp!
+$myhostname = host
```
kkaempf: No, that's not a `hostname`, that's a url.

MalloZup: hi @kkaempf ! thx for review! for moment i'm trying to running the suite, when i'm done i will make more order to the variables, also for the master branch :)

PR #51, opened 2016-08-03
Add test for minion default channel

mbologna:
A minion should always have a default base channel associated.
https://github.com/SUSE/spacewalk/pull/715

### DIFF
--- a/features/salt_minion_details.feature
+++ b/features/salt_minion_details.feature
@@ -19,7 +19,7 @@ Feature: Verify the minion registration
     Then I should see a "aaa_base" text
     And I should see a "aaa_base-extras" text
 
-  Scenario: Accepted minion have a base channel
+  Scenario: Accepted minion has a base channel
     Given that this minion is registered in Spacewalk
     And I follow this client link
     And I follow "Software"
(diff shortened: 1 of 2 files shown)


features/salt_minion_details.feature
```diff
@@ -15,3 +18,10 @@ Feature: Verify the minion registration
     And I follow "List / Remove"
     Then I should see a "aaa_base" text
     And I should see a "aaa_base-extras" text
+
+  Scenario: Accepted minion have a base channel
```
moio: Grammar nitpick: "have" is plural :smiling_imp:

mbologna:
@mcalmer @moio @mseidl this BDD test added works by assuming that mgr-sync has populated the related products.
Obviously, if database does not contain any product, minion should not have an associated channel.

This brings me to the question: do we want to sophisticate more the test to check if database products are added or is not necessary?

moio:
We execute tests in order, so we control when to run this test (before or after `mgr-sync refresh`).

To me this test is sufficient - it does not make too much sense to ensure it does not work before `mgr-sync refresh`. I would leave negative testing for jUnit.

MalloZup:
the test "have" a nice style,  for me :)

PR #52, opened 2016-08-18
Prevent multiple login failures because "Background" steps

meaksh:
This PR fixes:
- Scenario: Check for the Salt entitlement
- Scenario: Check installed packages are visible

We don't want to do a common login in "Background" for all scenarios because step `I am on the Systems overview page of this client` also handle the login for those tests. If multiple logins are done then tests will fail.

/cc @hustodemon @mseidl


mseidl:
this looks fine.

PR #53, opened 2016-08-19
Check if package list is refreshed for the accepted minion

meaksh:
This PR adds a new scenario to the `Verify minion registration` feature:

1. Scenario: Check if package list has been refreshed

This should also fix a timing issue refreshing the package list that produces test failures in the next scenario if the refreshing has not been completed.

Changes are for `master` and `manager30` branches.

/cc @mseidl @hustodemon


mseidl:
lgtm

meaksh:
Merged in `master` and `manager30`

PR #54, opened 2016-08-24
Fix: waiting and refreshing page until pending minion appears

meaksh:
This PR modifies the steps of:

1. Scenario: Minion is visible in the Pending section

Sometimes fingerprint hasn't appeared yet when test is loaded. Now, test is waiting and refreshing the page until it contains `pending` text.

This changes should be also merged in `master` branch.

/cc @mseidl


meaksh:
Merged in manager30 and master

PR #60, opened 2016-09-26
Added 'pkgset' beacon tests for SLES and RHEL minions (slenkins branch)

meaksh:
This PR adds tests for the `pkgset` beacon for SLES and RHEL minions using SLEnkins:

1. Scenario: Manually removing a package in a minion 
2. Scenario: Manually installing a package in a minion

When users manually perform package installation or removal directly in the minion (using `zypper` or `yum` commands) then the `pkgset` beacon is aware of these changes and does report them to the SUMA UI.

So package changes manually done by `zypper` or `yum` actions are shown in the UI.

/cc @MalloZup @mseidl

### DIFF
--- a/features/salt_pkgset_beacon.feature
+++ b/features/salt_pkgset_beacon.feature
@@ -1,0 +1,24 @@
+# Copyright (c) 2016 SUSE LLC
+# Licensed under the terms of the MIT license.
+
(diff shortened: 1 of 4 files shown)


MalloZup:
yop, cool, for me seems nice, i will test it

PR #63, opened 2016-10-28
Enable MINION-BOOTSTRAP.feature testing

MalloZup:


features/ssh_push.feature
```diff
@@ -38,3 +43,23 @@ Feature: Register a system to be managed via SSH push
...
+
+  # Salt minion bootstrap
+  Scenario: bootstrap a sles minion via ssh-push
+     Given I am authorized
+     When I follow "Salt"
+     Then I should see a "Bootstrapping" text
+     And I follow "Bootstrapping"
+     Then I should see a "Bootstrap Minions" text
+     # FIXME: we should maybe add the minion_ip in server/hosts, so we address systems without-ip this for redhat or sles.
+     And  I enter "#{$minion_ip}" as "hostname"
```
MalloZup: here as mentioned, we should maybe modify the /etc/hosts of the server, so that we can put the name of the minion, and not the ip.

features/ssh_push.feature
```diff
@@ -1,11 +1,16 @@
-# Copyright (c) 2015 SUSE LLC
+# Copyright (c) 2016 SUSE LLC
 # Licensed under the terms of the MIT license.
 
-Feature: Register a system to be managed via SSH push
+Feature: Register a system(traditional and salt) to be managed via SSH push
```
MalloZup: i thought that the feature was ok in the same traditional "ssh-push" feature, and not creating a new feature.

features/minion_ssh_push.feature
```diff
@@ -0,0 +1,23 @@
...
+     Then I should see a "Bootstrapping" text
+     And I follow "Bootstrapping"
+     Then I should see a "Bootstrap Minions" text
+     # minion = sles, rh_minion = redhat
+     And  I enter the hostname of "sle-minion" as hostname
+     And I enter "22" as "port"
+     And I enter "root" as "user"
+     And I enter "linux" as "password"
+     And I click on "Bootstrap it"
+     Then i should see a "Successfully bootstrapped host! Your system should appear in System Overview shortly." text
```
MalloZup: HERE maybe we need to wait until the minion is bootstrapped. We should check that same process still exist on the command line, then wer are done, to wait. ( don't know which process could be.)

features/minion_ssh_push.feature
```diff
@@ -0,0 +1,26 @@
...
+     Then I should see a "Bootstrapping" text
+     And I follow "Bootstrapping"
+     Then I should see a "Bootstrap Minions" text
+     # sle-minion = sles, rh_minion = redhat
+     And  I enter the hostname of "sle-minion" as hostname
+     And I enter "22" as "port"
+     And I enter "root" as "user"
+     And I enter "linux" as "password"
+     And I click on "Bootstrap it"
+     Then take picture
```
mbologna: I think it is enough just to use "I should see" like you do in the next line

features/step_definitions/spacewalk_errors.rb
```diff
@@ -3,5 +3,6 @@
 
 Then(/^I control that up2date logs on client under test contains no Traceback error$/) do
   cmd = "grep \"Traceback\" /var/log/up2date"
-  $client.run(cmd)
+  code, _out = $client.run(cmd)
+  raise "error found, check the client up2dat logs" if code == 0
```
mbologna: nitpick: typo "up2date"

features/step_definitions/minion_ssh_push_steps.rb
```diff
@@ -0,0 +1,50 @@
...
+            puts "Took #{finished.to_i - start.to_i} seconds to contact the minion"
+            break
+          end
+
+        elsif minion == "rh-minion"
+          _code, out = $server.run("salt #{$rh_minion_hostname} test.ping", false)
+          if out.include?($minion_hostname)
+            finished = Time.now
+            puts "Took #{finished.to_i - start.to_i} seconds to contact the minion"
+            break
```
mbologna: This block could be refactored for 'sle-minion' and 'rh-minion'. What do you think?

MalloZup: true, i had that in my todo list :)

features/minion_ssh_push.feature
MalloZup: yeah, for moment, i put take a picture just for testing, I wlll put then  "i should see " or something like this, that take care of waiting that the minion is bootstrapped.

features/step_definitions/spacewalk_errors.rb
MalloZup: fixed

PR #73, opened 2016-11-16
BUGFIXES: use O for wget, use http

MalloZup:


MalloZup:
Now only 1 test failing:D

PR #77, opened 2016-11-18
Total of available public api calls set to '675'

meaksh:
This PR set the total available public API calls to `675`.

We introduced a new method 'getInstalledProducts' (https://github.com/SUSE/spacewalk/pull/809) which is now returned by the `getApiNamespaceCallList` call. We need to increase the expected number of methods returned by this call to fix a failing test in our testsuite.

@cbbayburt - Are you agreed with this changes?

### DIFF
--- a/features/xmlrpc_api.feature
+++ b/features/xmlrpc_api.feature
@@ -8,5 +8,5 @@ Feature: Test XML-RPC "api" namespace.
     When I call getVersion, I should get "17" as result
     And I call systemVersion, I should get "3.0" as result
     And I call getApiNamespaces, I should get 48 namespaces
-    And I call getApiNamespaceCallList, I should get 674 available api calls
+    And I call getApiNamespaceCallList, I should get 675 available api calls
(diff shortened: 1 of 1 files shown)


MalloZup:
thx @meakshi!!  I  already fixed the new 665 calls for the new future master branch, but i didn't for the actual master branch. LGTM for me

meaksh:
@MalloZup - the new call is covered with unit tests. @cbbayburt do you think we need also a cucumber test for the new call or it's enough with unit tests?

MalloZup:
@meaksh @cbbayburt , if needed i can do the test in cucumber, just point me the code of the new call. If you think that the unit-test is enough, we can just merge then. :)

cbbayburt:
@meaksh, @MalloZup, `getInstalledProducts` has a unit test, and I think that is enough. So, no need for a cucumber test for this one.

meaksh:
@cbbayburt - Great! Then I'll merge this PR. Thanks!

PR #139, opened 2016-12-20
Enable salt_ssh feature

MalloZup:


MalloZup:
@mateiw , what do you think

PR #175, opened 2017-01-30
wrong variable name _code => $fail_code

Bischoff:
first case was using $fail_code, others were using _code

switch ... when instead of if ... elsif

better English

### DIFF
--- a/features/step_definitions/command_steps.rb
+++ b/features/step_definitions/command_steps.rb
@@ -217,20 +217,23 @@ When(/^I take a snapshot "([^"]*)"$/) do |name|
 end
 
 When(/^I run "([^"]*)" on "([^"]*)"$/) do |cmd, target|
-  if target == "server"
+  case target
+  when "server"
     _out, $fail_code = $server.run(cmd, false)
-  elsif target == "ceos-minion"
-    _out, _code = $ceos_minion.run(cmd, false)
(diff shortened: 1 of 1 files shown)


features/step_definitions/command_steps.rb
```diff
@@ -217,20 +217,21 @@ def checkRestart(host, time_out)
...
-  elsif target == "sle-client"
-    _out, _code = $client.run(cmd, false)
+  when "ceos-minion"
+    _out, $fail_code = $ceos_minion.run(cmd, false)
+  when "ssh-minion"
+    _out, $fail_code = $ssh_minion.run(cmd, false)
+  when "sle-minion"
+    _out, $fail_code = $minion.run(cmd, false)
+  when "sle-client"
+    _out, $fail_code = $client.run(cmd, false)
```
MalloZup: we can add here a else statement 

``` else 
 raise "no valid target given"```
for be in the safe side.

It look goods for me, just rebase then after the 3 commits in one

Bischoff: yup.

MalloZup:
@Bischoff , we need also to backport this change to the 3.0 theoretically, if the step is there.

MalloZup:
ok rebase the 3 commits in one and i merge it

PR #176, opened 2017-01-30
New feature: migrate a traditional client into a salt minion

Bischoff:
Kind of minimalistic for the time being. Tests for bsc#1020902, but should also test many other aspects of the minion are like before the migration.

Ideally, we would also migrate a traditional Centos client, especially for bsc#1012956. But that would mean one more VM for one test only... :-(

### DIFF
--- a/features/migrate_client_to_minion.feature
+++ b/features/migrate_client_to_minion.feature
@@ -1,0 +1,30 @@
+# Copyright (c) 2017 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Migrate a traditional client into a salt minion
+
+  Scenario: Migrate a sles client into a salt minion
+     Given I am authorized
+     When I follow "Salt"
+     Then I should see a "Bootstrapping" text
+     And I follow "Bootstrapping"
+     Then I should see a "Bootstrap Minions" text
+     # sle-client = traditional sles client
+     And I enter the hostname of "sle-client" as hostname
+     And I enter "22" as "port"
+     And I enter "root" as "user"
+     And I enter "linux" as "password"
+     And I select "1-SUSE-PKG-x86_64" from "activationKeys"
+     And I click on "Bootstrap"
+     And I wait for "100" seconds
+     Then I should see a "Successfully bootstrapped host! Your system should appear in System Overview shortly." text
+
+  Scenario: Verify the minion was bootstrapped with activation key
+     Given I am on the Systems overview page of this "sle-client"
+     Then I should see a "Activation Key: 	1-SUSE-PKG" text
+
+  Scenario: Check that service nhsd has been stopped
+     # bsc#1020902 - moving from traditional to salt with bootstrap is not disabling rhnsd
+     When I run "systemctl status nhsd | grep Active" on "sle-client"
+     Then the command should fail
+
--- a/features/step_definitions/minion_bootstrap_steps.rb
+++ b/features/step_definitions/minion_bootstrap_steps.rb
@@ -2,13 +2,15 @@
 # Licensed under the terms of the MIT license.
 
 And(/^I enter the hostname of "([^"]*)" as hostname$/) do |minion|
-  if minion == "sle-minion"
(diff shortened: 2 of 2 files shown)


features/migrate_client_to_minion.feature
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2017 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Migrate a traditional client into a salt minion
+
+  Scenario: Migrate a sles client into a salt minion
+     Given I am authorized
+     When I follow "Salt"
+     Then I should see a "Bootstrapping" text
```
MalloZup: You can remove the" I should see a bootstrapping text"

features/migrate_client_to_minion.feature
```diff
@@ -0,0 +1,30 @@
...
+     And I follow "Bootstrapping"
+     Then I should see a "Bootstrap Minions" text
```
MalloZup: the ``` then ```we can remove it, keep it simple, since we are not testing the page, we want only to boostrap

features/migrate_client_to_minion.feature
```diff
@@ -0,0 +1,30 @@
...
+     And I enter "22" as "port"
+     And I enter "root" as "user"
+     And I enter "linux" as "password"
+     And I select "1-SUSE-PKG-x86_64" from "activationKeys"
+     And I click on "Bootstrap"
+     And I wait for "100" seconds
+     Then I should see a "Successfully bootstrapped host! Your system should appear in System Overview shortly." text
+
+  Scenario: Verify the minion was bootstrapped with activation key
+     Given I am on the Systems overview page of this "sle-client"
```
MalloZup: the "sle-client" here, is kind of hack, because now is a sle-minion. We schould add a variable in the switch statement,  "sle-migrated-minion", but the target will be the sle-client. (if we go further with test, we maybe consider to think further), but for this specific feature is ok for me

features/migrate_client_to_minion.feature
```diff
@@ -0,0 +1,30 @@
...
+     Then I should see a "Activation Key: 	1-SUSE-PKG" text
+
+  Scenario: Check that service nhsd has been stopped
+     # bsc#1020902 - moving from traditional to salt with bootstrap is not disabling rhnsd
+     When I run "systemctl status nhsd | grep Active" on "sle-client"
```
MalloZup: same as above, replace "sle-client" with a migration name 👯‍♂️

MalloZup:
@Bischoff, good work so far !  I like your approach. My remarks are in the code.

For the centos_7 feature, we don't need a new-machine:
The centos7 vm can initialized as tradional-client, "migrated as a salt-client", "deleted", boostrapped as salt-client ( this last step we cover already)
But in my view this need a separate card, i would concentrate on this feature.

features/migrate_client_to_minion.feature
Bischoff: As for reusing the same VM for centos7 (bug 1012956), it's a nice idea. However, this creates a problem: you will need to migrate it FIRST, then do the normal tests with the resulting minion. With the SLE12, we currently do all tests, then migrate it AFTER. We can have two different cucumber features at two different points of time, but I don't like the idea, I'd prefer to group all tests of migration in one single feature. I have no good solution at this point.

As for renaming sle-client, if you don't mind I'll do in next days a big cleanup of these:
- good English sle-client => SLE client
- group all switch / if... elsif statements in one ruby function
and I'll be able to add the "SLE migrated" one as you request.

MalloZup: @Bischoff , i would like that you add it now"SLE migrated", the rest you can do after, you have my benediction for that.

Concerning the centos7, we have already feature like cento_salt_test: at beginning i was thinking like you, but with experience, is a good practice to have different feature even for different os-family. for debugging and tracing bug is much more better

Bischoff:
continuing in a sparate pull request (180)

PR #182, opened 2017-01-31
Fixed a mistake I introduced in last PR (sorry)

Bischoff:
This was supposed to be a fix, but generated an error...


MalloZup:
Trad client use not full hostname in gui

PR #183, opened 2017-02-01
Fixed test for active process

Bischoff:
systemctl status" displays "Active" in all cases,
it was a bad idea to test for it.


Bischoff:
already done by Dario

PR #184, opened 2017-02-02
Restart phantomjs browser at begin of each feature

MalloZup:
We avoid ram issue in this way.


MalloZup:
restesting last time the fix, i tested it already one time

PR #187, opened 2017-02-03
Test scenarios for migration from classical client to Salt minion

Bischoff:
These tests test migration from traditional client to Salt minion.

They have been tried in isolation and do work in such conditions.

They DO FAIL when run inside of the whole test suite because of product bug bsc#1023399 - "Migrate a traditional client in SSH push mode into a salt minion doesn't work", but that is expected.

### DIFF
--- a/features/migrate_client_to_minion.feature
+++ b/features/migrate_client_to_minion.feature
@@ -70,8 +70,9 @@ Feature: Migrate a traditional client into a salt minion
      And I enter as remote command this script in
       """
       #!/bin/bash
-      touch /tmp/tastes-better-with-salt
+      touch /tmp/remote-command-on-migrated-text
       """
      And I click on "Schedule"
      Then I should see a "Remote Command has been scheduled successfully" text
-     And "/tmp/tastes-better-with-salt" exists on the filesystem
+     And "/tmp/remote-command-on-migrated-test" exists on the filesystem of "sle-migrated-minion"
+     And I remove "/tmp/remote-command-on-migrated-test" from "sle-migrated-minion"
--- a/features/migrate_client_to_minion.feature
+++ b/features/migrate_client_to_minion.feature
@@ -2,6 +2,9 @@
 # Licensed under the terms of the MIT license.
 
 Feature: Migrate a traditional client into a salt minion
+  In order to move away from traditional clients
+  As an authorized user
+  I want to migrate these clients to salt minions and have everything as before
 
   Scenario: Migrate a sles client into a salt minion
      Given I am authorized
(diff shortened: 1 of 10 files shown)


features/migrate_client_to_minion.feature
```diff
@@ -2,6 +2,9 @@
 # Licensed under the terms of the MIT license.
 
 Feature: Migrate a traditional client into a salt minion
+  In order to move away from traditional clients
```
MalloZup: i think we can remove this line. the two after are enough and describe better the feature.

features/migrate_client_to_minion.feature
```diff
@@ -25,3 +28,50 @@ Feature: Migrate a traditional client into a salt minion
...
+     When I follow "Software" in the content area
+     And I follow "Install"
+     And I check "perseus-dummy-1.1-1.1" in the list
+     And I click on "Install Selected Packages"
+     And I click on "Confirm"
+     And I wait for "5" seconds
+     Then I should see a "1 package install has been scheduled for" text
+     And I wait for "perseus-dummy-1.1-1.1" to be installed
+
+  Scenario: Run a remote command on the migrated minion
```
MalloZup: can you remove the test file you generated in this scenario. At the end of scenario

features/migrate_client_to_minion.feature
```diff
@@ -25,3 +28,50 @@ Feature: Migrate a traditional client into a salt minion
...
+     Given I am on the Systems overview page of this "sle-migrated-minion"
+     When I follow "Remote Command" in the content area
+     And I enter as remote command this script in
+      """
+      #!/bin/bash
+      touch /tmp/tastes-better-with-salt
+      """
+     And I click on "Schedule"
+     Then I should see a "Remote Command has been scheduled successfully" text
+     And "/tmp/tastes-better-with-salt" exists on the filesystem
```
MalloZup: I don't like the step ``` And "/tmp/tastes-better-with-salt" exists on the filesystem```.

For **1** simply reason: which **filesystem** are we addressing here ``minion, server, etc``? It's not really clear and we should avoid this kind of hardcoded steps.

I think you should just use the step I run ```test -f $myfile``` on "minion" and then ```rm ``` on ```target```

MalloZup: I saw the implementation here ```features/step_definitions/salt_steps.rb``, @ebischoff, if you want to eliminate the step and make it better, for me is ok. otherwise keep the old one just because it works, and use a new correct one, as you wish

MalloZup:
```
features/step_definitions/salt_steps.rb:When(/^"(.*)" exists on the filesystem$/) do |file|
features/action_chain.feature:    Then "/root/webui-actionchain-test" exists on the filesystem
features/salt_remote_cmds.feature:    And "/root/12345" exists on the filesystem
```

features/migrate_client_to_minion.feature
Bischoff: No, it describes the context. Someone less familiar with the testsuite would find this information of value.

features/migrate_client_to_minion.feature
Bischoff: OK. Will do.

features/migrate_client_to_minion.feature
Bischoff: Agree with the approach "filesystem of ...".

features/salt_states_catalog.feature
```diff
@@ -42,4 +42,4 @@ Feature: Check the Salt package state UI
     Then I should see a "1 Changes" text
     And I click on the css "button#save-btn"
     And I click on the css "button#apply-btn"
-    Then I wait for the file "/root/foobar"
+    Then "/root/foobar" exists on the filesystem of "sle-minion"
```
MalloZup: "I wait for the file" was another function that did something else, why do you removed and replace the function?

features/step_definitions/command_steps.rb
```diff
@@ -216,28 +216,34 @@ def checkRestart(host, time_out)
...
+  _out, $fail_code = node.run(cmd, false)
 end
 Then(/^the command should fail$/) do
    raise "Previous command must fail, but has NOT failed!" if $fail_code.zero?
 end
 
+When(/^"(.*)" exists on the filesystem of "(.*)"$/) do |file, host|
+  node = get_target(host)
+  begin
+    Timeout.timeout(DEFAULT_TIMEOUT) do
```
MalloZup: ok, imho why we conserved here the loop with this complex structure ?

I think we can remove this at all.  

The function is ``` "abc" exists on the filesystem of X-target```, so we can assume that we don't wait for the creation of the file.

We could have simply a call and simplify this function a lot

MalloZup:
@Bischoff , great PR.

A lot of improvements.

- [ ] Review the changes and make the fixes
- [ ] Test the PR with whole suite
- [ ] Squash commits, or rename the commits with appropriate name,  ``` More test scenarios,  	Applied requested changes and fixed bug, Make cucumber_lint happy``` are not conform with our guideline *** How to Contribute```

https://github.com/SUSE/spacewalk-testsuite-base#0-make-a-good-commit-message

features/salt_states_catalog.feature
Bischoff: Yes I did. There were two functions doing almost the same thing:
 * wait for the file => timeout == 60 seconds
 * file exists => timeout == 200 seconds

features/step_definitions/command_steps.rb
Bischoff: No. It would probably lead to failures, as some file creations are asynchronous (for example through action chains).

What we could do is either do a function with no timeout for the synchronous cases (if any), or do a function with smaller timeout or a parametrized timeout. But none of this should be done without previous accurate analysis.

PR #189, opened 2017-02-07
[WIP:]Backport from 3.1: migration tests, fixes

Bischoff:
Tests backported from HEAD branch:
 * new migration tests
 * tell which system we check for a file
 * fixed bug of file_exist() that was always returning true


MalloZup:
After test rebase commit

PR #192, opened 2017-02-17
[WIP] Enabling to move the migration feature to another point in test suite

Bischoff:
Before Ornela writes a test that would test both minion and ssh push minion, I made the code better so it would run even from an earlier point in the test suite, so we have more green.

While doing it I found bugs in my code. These fixes are also included.

### DIFF
--- a/features/migrate_client_to_minion.feature
+++ b/features/migrate_client_to_minion.feature
@@ -30,8 +30,6 @@ Feature: Migrate a traditional client into a salt minion
      Then the command should fail
 
   Scenario: Check that the migrated system is now a minion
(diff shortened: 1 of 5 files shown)


features/migrate_client_to_minion.feature
```diff
@@ -30,28 +30,21 @@ Feature: Migrate a traditional client into a salt minion
      Then the command should fail
 
   Scenario: Check that the migrated system is now a minion
+     # we should also try to migrate a traditional client in SSH push mode because of
```
MalloZup: Remove the comment and put this in a card or in an issue. 

Such comments will remain forever otherwise

PR #195, opened 2017-03-02
Fix testsuite to use the tristate button

MalloZup:


MalloZup:
me neither ! but so it's life. there is no simply solution for that 💃

PR #196, opened 2017-03-03
Manager-31: Init docker feature tests

MalloZup:


features/docker_build_image.feature
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2017 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature:  Build Container images with SUSE Manager
+  
+  Scenario: Transform a salt-minion to a Container Build Host
+  Scenario: Create an Image Store
```
mcalmer: we have authenticated and not authenticated. So we could have tests for both.

features/docker_build_image.feature
```diff
@@ -0,0 +1,10 @@
...
+  Scenario: Create an Activation Key to define the used channels during building
+  Scenario: Create an Image Profile
+  Scenario: Build a docker Image
```
mcalmer: we need some more tests for checking the results of the image build.

MalloZup: yeah, i agree. for moment was just an init for the skeletron :)

PR #209, opened 2017-03-09
Testing Salt formulas

Bischoff:
Tested both in isolation and within the test suite with success.

### DIFF
--- a/features/step_definitions/salt_steps.rb
+++ b/features/step_definitions/salt_steps.rb
@@ -1,4 +1,4 @@
-# Copyright 2015-16 SUSE LLC
+# Copyright 2015-2017 SUSE LLC
 require 'timeout'
 
 Given(/^the Salt Minion is configured$/) do
@@ -233,24 +233,6 @@ Then(/^all local repositories are disabled$/) do
     end
 end
 
-Then(/^I try to reload page until contains "([^"]*)" text$/) do |arg1|
-  found = false
-  begin
-    Timeout.timeout(30) do
-      loop do
-        if page.has_content?(debrand_string(arg1))
-          found = true
-          break
-        end
-        visit current_url
-      end
-    end
-  rescue Timeout::Error
-    raise "'#{arg1}' cannot be found after wait and reload page"
-  end
-  fail unless found
-end
-
 And(/^I follow the sle minion$/) do
  step %(I follow "#{$minion_fullhostname}")
 end
@@ -317,7 +299,6 @@ Then(/^I should see "([^"]*)" in the command output$/) do |text|
   end
 end
 
-# salt_pkgset_beacon_steps.rb
 Then(/^I manually install the "([^"]*)" package in the minion$/) do |package|
   if file_exists?($minion, "/usr/bin/zypper")
     cmd = "zypper --non-interactive install -y #{package}"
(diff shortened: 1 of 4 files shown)


Bischoff:
Yes, what you suggest makes sense.

But if I have a look at the salt steps, I can see

* features/step_definitions/salt_download_endpoint_steps.rb
* features/step_definitions/salt_formulas_steps.rb
* features/step_definitions/salt_minions_page_steps.rb
* features/step_definitions/salt_software_states.rb
* features/step_definitions/salt_ssh_steps.rb
* features/step_definitions/salt_steps.rb

so it is already split across several files. Also, there is the risk to have a salt steps file that is too big.

I suggest we discuss this calmly after the GMC, as part of the many cleanups. Meanwhile, I'd suggest it can stay like this.

I'll add it to the "cleanup" card in Trello.

MalloZup:
NO, please, no new errors. the past is the past, the present is to improve.

put the steps here ```features/step_definitions/salt_steps.rb```

Bischoff:
Question is, is it really an improvement?

Having better classified steps has also its advantages.

MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base#1-steps-should-be-grouped-by-topic-and-not-by-feature

This guideline was made by @dmacvicar 

https://github.com/SUSE/spacewalk-testsuite-base/blob/master/docs/Pitfalls-test.md#meanless-step-names

We are not at the level but we want to go on the direction

Bischoff:
As you wish.

I'll move them to `salt_steps.rb` tomorrow.

MalloZup:
Éric no hurry take your time i reccomend you also the guide link i gave you it à nice doc

Bischoff:
Done :-) .

features/step_definitions/salt_steps.rb
```diff
@@ -233,6 +233,10 @@
     end
 end
 
+Then(/^I reload the page$/) do
```
MalloZup: @Bischoff  this should go into ``` navigation_steps```

features/step_definitions/salt_steps.rb
```diff
@@ -376,3 +379,65 @@
...
+      "timezone\$name"
+    when "language"
+      "keyboard_and_language\$language"
+    when "keyboard layout"
+      "keyboard_and_language\$keyboard_layout"
+  end
+  select(value, :from => boxid)
+end
+
+Then(/^the timezone on the minion should be "([^"]*)"$/) do |timezone|
```
MalloZup: we should use here sle-minion in sake of readibility.
Because $minion is sle-minion, otherwise for super-generic you should use the " " for taking input minion. But we don't need super-generic there

features/step_definitions/salt_steps.rb
```diff
@@ -376,3 +379,65 @@
...
+  output, _code = $minion.run("date +%Z")
+  fail unless output.strip == timezone
+end
+
+Then(/^the keymap on the minion should be "([^"]*)"$/) do |keymap|
```
MalloZup: same here, imho "sle-minion" is much more consistent"

features/step_definitions/salt_steps.rb
```diff
@@ -376,3 +379,65 @@
...
+  output, _code = $minion.run("cat /etc/vconsole.conf")
+  fail unless output.strip == "KEYMAP=#{keymap}"
+end
+
+Then(/^the language on the minion should be "([^"]*)"$/) do |language|
```
MalloZup: same here

features/step_definitions/salt_steps.rb
Bischoff: OK. But "sle-minion" is not English. It should become "SLE minion" at the end everywhere.

features/step_definitions/salt_steps.rb
Bischoff: ok.

Bischoff: but so should the "I wait for reload" just below, right?

Bischoff:
Latest change requests applied, retested in isolation -> green. Tests in context not needed as only moving around and changing cucumber stances.

MalloZup:
OK looks good thx!  We will merge it after beta1

Bischoff:
Thank you Dario.

MalloZup:
Yw

PR #213, opened 2017-03-14
Apply highstate for container host

MalloZup:


MalloZup:
need merge before of this:
https://github.com/moio/sumaform/pull/96

MalloZup:
@MC where we should get the python docker pkg>1.6? We have only i à 1,5 pkg on container repos

mc:
No idea....

 -mc

PR #220, opened 2017-03-16
Enable dockerstate in testsuite

MalloZup:


MalloZup:
this is were added in sumaform, not need anymore for the state

PR #228, opened 2017-03-21
WIP: adding container with server host and inspect scenario prototype

MalloZup:


features/step_definitions/docker_steps.rb
```diff
@@ -39,6 +39,16 @@ def retrieve_minion_id
   visit("https://#{$server_fullhostname}/rhn/manager/cm/build")
 end
 
+And(/^I delete the image "([^"]*)" via xmlrpc-call$/) do |imagetodel|
+  cont_op.login('admin', 'admin')
+  # get the image id
+  images_list = []
+  images_list = cont_op.listImages("admin")
+  refute_nil(images_list, "ERROR: no images were retrieved.")
+  imageid = images_list.select { |imagelisted| imagelisted == imagetodel }
```
MalloZup: @mcalmer do you know how can i get the imageid of an image?
I looked at the methods on xmlrpc but i found out that the only image.scheduleBuild method returned  a Build ID.

The image id is this build id? or where we can retrieve it in order to pass it?

cbbayburt: `listImages` should return the IDs actually. On the other hand, `scheduleBuild` returns the action ID of the action which is scheduled on the build host for building the image.

MalloZup:
i splitted the pr in small prs.. 🖤

Issue #231, opened 2017-03-22
I run on ** steps make wrong assumptions


MalloZup:
``` 
When(/^I run "([^"]*)" on "([^"]*)"$/) do |cmd, host|
  node = get_target(host)
  _out, $fail_code = node.run(cmd, false)
end
```

This step should be renamed and we need to split in two. This steps doesn't check the failure of the command , so if used, we maybe expect that if the command fail the steps fail, but isn't it

PR #232, opened 2017-03-22
Fix broken step

MalloZup:
fix https://github.com/SUSE/spacewalk-testsuite-base/issues/231


MalloZup:
the only step was using the "without error control" where the steps with "it should fail"

PR #233, opened 2017-03-22
Added tests for openSCAP audits on salt minions

Bischoff:
Tests:
* audit salt minions with openscap (currently fails, product bug: permissions problem)
* compare two audit reports
* delete audit reports (currently fails, product bug)

Tested both in isolation and within whole testsuite.

### DIFF
--- a/features/salt_openscap_audit.feature
+++ b/features/salt_openscap_audit.feature
@@ -1,0 +1,59 @@
+# Copyright (c) 2017 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Use the openSCAP audit feature in SUSE Manager for a salt minion
+  In order to audit my salt minions
+  As an authorized user
+  I want to run openSCAP scans on them
+
+  Scenario: Schedule an audit job on the minion
+    Given I disable IPv6 forwarding on all interfaces of the SLE minion
+    And I am on the Systems overview page of this "sle-minion"
+    And I follow "Audit" in the content area
+    And I follow "Schedule" in the content area
+    When I enter "--profile Default" as "params"
+    And I enter "/usr/share/openscap/scap-yast2sec-xccdf.xml" as "path"
+    And I click on "Schedule"
+    Then I should see a "XCCDF scan has been scheduled" text
+    And I wait for the openSCAP audit to finish
+
+  Scenario: Check results of the audit job on the minion
+    Given I am on the Systems overview page of this "sle-minion"
+    And I follow "Audit" in the content area
+    When I follow "xccdf_org.open-scap_testresult_Default"
+    Then I should see a "Details of XCCDF Scan" text
+    And I should see a "Default" text
+    And I should see a "XCCDF Rule Results" text
+    And I should see a "pass" text
+    And I should see a "rule-" link
+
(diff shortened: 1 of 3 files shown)


features/salt_openscap_audit.feature
```diff
@@ -0,0 +1,58 @@
...
+
+  Scenario: Delete audit results
+    Given I am on the Systems overview page of this "sle-minion"
+    And I follow "Audit" in the content area
+    And I follow "List Scans" in the content area
+    When I click on "Select All"
+    And I click on "Remove Selected Scans"
+    And I click on "Confirm"
+    Then I should see a "2 SCAP Scan(s) deleted. 0 SCAP Scan(s) retained" text
+
```
MalloZup: Disable ipv6 after test, so make the minion as it was before

features/step_definitions/salt_steps.rb
```diff
@@ -442,3 +442,11 @@
   output, _code = $server.run("salt '#{$minion_ip}' pillar.get '#{key}'")
   fail unless output.split("\n")[1].strip == value
 end
+
+When(/^I disable IPv6 forwarding on all interfaces of this "sle-minion"$/) do
```
MalloZup: do we need here "sle-minion" ?

We can remove because it's a specific steps so we can just use 

```I disable IPv6 forwarding on all interfaces of the sle-minion/```

The "sle-minion" we use this for suggest that is a variable.

File: features/step_definitions/salt_steps.rb [hunk 3]
```diff
@@ -442,3 +442,11 @@
...
+  $minion.run("sysctl net.ipv6.conf.all.forwarding=0")
+end
+
+When(/^I enable IPv6 forwarding on all interfaces of this "sle-minion"$/) do
```
MalloZup: same as above

features/salt_openscap_audit.feature
```diff
@@ -0,0 +1,58 @@
...
+  Scenario: Schedule an audit job on the minion
+    Given I am on the Systems overview page of this "sle-minion"
+    And I disable IPv6 forwarding on all interfaces of this "sle-minion"
+    And I follow "Audit" in the content area
+    And I follow "Schedule" in the content area
+    When I enter "--profile Default" as "params"
+    And I enter "/usr/share/openscap/scap-yast2sec-xccdf.xml" as "path"
+    And I click on "Schedule"
+    Then I should see a "XCCDF scan has been scheduled" text
+    And I wait for "10" seconds
```
MalloZup: @Bischoff  which action are we waiting here for?
In order to avoid the usage of sleep

Bischoff: the end of audit scan - I might be able to test for openscap process...

features/salt_openscap_audit.feature
Bischoff: done

ideally, i would remember the value before in a variable, instead of assuming it's zero

features/step_definitions/salt_steps.rb
Bischoff: done ("of the minion")

MalloZup:
@Bischoff  great! please remove my name on the commit, 🖤  and squash them

PR #240, opened 2017-03-25
Setting scans retention period to zero, so scans can be deleted

Bischoff:
Solves testsuite failure when removing openscap scans

### DIFF
--- a/features/salt_openscap_audit.feature
+++ b/features/salt_openscap_audit.feature
@@ -37,6 +37,7 @@ Feature: Use the openSCAP audit feature in SUSE Manager for a salt minion
     And I click on "Schedule"
     Then I should see a "XCCDF scan has been scheduled" text
     And I wait for the openSCAP audit to finish
+    And I disable IPv6 forwarding on all interfaces of the SLE minion
 
   Scenario: Compare audit results
     Given I am on the Systems overview page of this "sle-minion"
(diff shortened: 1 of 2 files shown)


features/salt_openscap_audit.feature
```diff
@@ -47,9 +48,16 @@ Feature: Use the openSCAP audit feature in SUSE Manager for a salt minion
...
+  Scenario: Remove audit scans retention period
+    Given I am on the Organizations page
+    And I follow "SUSE Test" in the content area
+    And I follow "Configuration" in the content area
+    When I enter "0" as "scap_retention_period"
+    And I click on "Update Organization"
+    Then I should see a "Organization SUSE Test was successfully updated." text
+
+  Scenario: Delete audit results
+    Given I am on the Systems overview page of this "sle-minion"
```
MalloZup: Add cleanup scenario to set the var to default

Bischoff:
True. Doing it now.

MalloZup:
@Bischoff looks good.

Can you rename this Scenario name to be more precise?

https://github.com/Bischoff/spacewalk-testsuite-base/blob/2a27e87a2f9dbbb77ea4d2c7ebcdd3406961c6b9/features/salt_openscap_audit.feature#L30

What is the intention/goal to have an quite identical audit?

Bischoff:
I'll merge the pull request because I need this in the test suite review for tomorrow.

About the name: I don't see how this is unprecise. Maybe the fact that it is "almost identical" is unrelevant and could be removed. It is almost identical because only one IPv6 variable change is detected.

MalloZup:
It is not clear. Why you make a second run identical then?

PR #247, opened 2017-03-31
Better action pickup scenario (still disabled)

Bischoff:
The fix puts the PID of the remote command into the test file name, to avoid race between several tests.

But the real problem is that the GUI does not sort the actions by chronological order, so picking up the "first" action is wrong. I don't see a solution to this unless we change the GUI.

These changes can safely be merged though as the cucumber feature is still commented out...


MalloZup:
@Bischoff Ok, thx for diving it. since it's commented i will merge it, so far you improved the steps so it's ok 🇭🇹

PR #251, opened 2017-03-31
Graphs for v2

MalloZup:


MalloZup:
just the preview how the new design should be of v2

MalloZup:
THIS PR Put everyting in v2 so the prototype can be merged to be more added, the reviewer is only for infos about the changes

PR #271, opened 2017-05-15
Manager30 rebootmachines

mseidl:
this is part of a sumaform / testsuite change.  Sumaform will (soon) patch machines with sle test updates which will require a reboot(from kernel/glibc).   So they will now be restarted in the test suite first, before the rest of the tests run.

### DIFF
--- a/features/reboot_machines.feature
+++ b/features/reboot_machines.feature
@@ -1,0 +1,7 @@
+# Copyright (c) 2017 SUSE LLC                                                                                                                             
+# Licensed under the terms of the MIT license. 
+
+Feature: I reboot the sles vms
+
+Scenario: Reboot sles vms
+  When I reboot sles vms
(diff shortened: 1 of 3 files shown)


MalloZup:
@mseidl  thx for the PR. 

I think however that we should fix/adress this reboot problem in sumaform:

-  this is not really a test.
- we boot always machines, even if we don't need the reboot ( when we don't use this flag, we reboot anyway)

mseidl:
Here is the issue in why we need it here... salt does a reboot and wait for it to come back up, but _only_ in a salt master setup, we are master-less.

MalloZup:
ok, then we can make a script, that will check the existence of grains, then reboot the machines it the grain is present.

MalloZup:
since was inactive, we will close this  and atm we have other priority, i will create an issue to readress this

PR #278, opened 2017-05-23
Manager 3.1: UI Changes

cbbayburt:
The changes introduced in the following pull requests:
- [Manager: Highstate UI](https://github.com/SUSE/spacewalk/pull/1145)
- [Manager: Minor UI Changes](https://github.com/SUSE/spacewalk/pull/1153)

### DIFF
--- a/features/baremetal_test.feature
+++ b/features/baremetal_test.feature
@@ -25,11 +25,11 @@ Feature: Test Bare-metal discovery
     And I should see a "Edit These Properties" link
     And I should not see a "[Management]" text
 
-  Scenario: see the client in Bare metal specific system list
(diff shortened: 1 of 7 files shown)


features/support/env.rb
```diff
@@ -50,6 +50,7 @@ def debrand_string(str)
         when "Create Kickstart Profile" then str
         when "Test Erratum" then str
         # replacement exceptions
+        when "Bare Metal Kickstart" then "Unprovisioned Autoinstallation"
```
MalloZup: we want to get ride of debrand string function , see on issues, for the historical idea. So we dont need the changes there, but directly on page

cbbayburt: oh, ok. let me change it.

PR #290, opened 2017-06-09
Add a step to wait for onboarding completed

mcalmer:
Not yet finished, but good as a first step

### DIFF
--- a/features/step_definitions/salt_steps.rb
+++ b/features/step_definitions/salt_steps.rb
@@ -164,8 +164,11 @@ When(/^we wait till Salt master sees this minion as rejected$/) do
     )
 end
 
-Then(/^we wait until onboarding is completed$/) do
+Then(/^I wait until onboarding is completed for "([^"]*)"$/) do |system|
   steps %(
+    When I follow "Systems" in the left menu
+    And I follow "Overview" in the left menu
+    And I follow this "#{system}" link
     When I follow "Events"
     And I follow "History"
     Then I try to reload page until contains "Package List Refresh scheduled by (none)" text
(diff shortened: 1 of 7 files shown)


features/step_definitions/salt_steps.rb
```diff
@@ -164,6 +164,17 @@
     )
 end
 
+Then(/^I wait until onboarding is completed for "([^"]*)"$/) do |system|
+  steps %(
+    When I navigate to "rhn/systems/Overview.do" page
+    And I follow this "#{system}" link
+    When I follow "Events"
+    And I follow "History"
+    Then I try to reload page until contains "Package List Refresh scheduled by (none)" text
```
MalloZup: thx @mcalmer for the PR! the only step that is missing, afaik  is that we  need to ensure that Package list refresh was done with success, because otherwise it could fail (for some issue or bugs we had in the past).

Imho , we can do that only on 2 ways:
- 1) via xmlrpc ( need more time but would be the stable one) ,  minus -> we don't test gui 
- 2)  via gui: plus we test gui, minus -> how much is stable, and we need to find a  stable way to check that 🤙 

We already at the problem with this similar kind of test, the test "action_pickedup", imho the gui is kind of not reliable at moment for this test. 

I would vote, for xmlrpc to check the status of the action, i think we need only 2 function for that:

1) retrieve action ID of the "Packace list refresh", with api: ```listInProgressActions```
   * maybe we can call it ```get_actionID_byname("Package List Refresh scheduled by (none)", "sle-minion )```
2) use then function `wait_action_complete`: https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/step_definitions/centos_tradclient.rb#L20

Issue #225, opened 2017-03-20
features not yet automated [BACKLOG]


MalloZup:
## NEW: NEED PRIOTIZATION
(Add here feature that you are not sure which prio need to be set)

- CentOS 7 Spacewalk Features – Software Crashes  ( should we/suse-mgr test that?

## HIGH
- proxy (Whole feature is missing) 
- visualization (whole feature is missing)
-  [salt-ssh] Minion: test that  run.cmd, pkg install, and patch install are ok
- trad-client centos7: test patch installation
- minion centos7 : test patch installation

## MEDIUM
- search a system based on its location (country, state/province and city)
- system groups ( we have only a test for creating a group)
- bootstrap script with option "bootstrap with salt"
- salt formulas on SSH minion
- reschedule failed state/highstate 
- Schedule highstate apply for multiples systems with SSM: `System group > States > Highstate`
  -> tests if minion one of the minions is down, test with other minions up

## LOW
- System Set Manager : better coverage
- Test Inspect status and Reinspect button & dialog in image overview page

## technical debt: 
( this test are not maintenable in current status, take to much time, make the whole testsuite instable more difficult to debug.)
- migrate a traditional client in SSH push mode into a salt minion
- migrate a RHEL traditional client into a salt minion, and make sure rhnplugin is disabled
- migrate from a service pack to the other with synchronized repos (SALT minion)
- migrate from a service pack to the other with synchronized repos (traditional client)
- migration tests : better coverage (more minion types - 1. trad>salt (implemented), 2. trad>salt-ssh-push,

Bischoff:
I can see above a link, but that doesn't tell me what the issue is, and once this is defined, why we should do that.

Bischoff:
Also, I don't understand the title of this issue. What is it about ?

MalloZup:
@Bischoff i was thinking that we conserve this issue, but maybe for future  we just add an issue for every feature we need to do ( assuming this becomes not to big)

Bischoff:
Personally, I prefer having a single list, there is more visibility.

It would even be better with a list in git, so we have history.

I'm not too found of these nifty web tools. I find this "issues" tool confused.

MalloZup:
@Bischoff ok, then we stay with one big-list.

MalloZup:
i added here like 10 peoples that are from all mixed team, gui, salt, etc, (git limited to 10), so you can assign/view/review/ the list

kkaempf:
[...]

PR #295, opened 2017-06-19
Checking if there is no formula leaks on pillar data (bsc#1044236)

meaksh:
This PR ensures that there is no leaks of formulas on the pillar data.

Spacewalk PR: https://github.com/SUSE/spacewalk/pull/1227

This fixes one item of **[HIGH]** on https://github.com/SUSE/spacewalk-testsuite-base/issues/225

Also I created a PR to `sumaform` as `salt-ssh` needs to be installed on the testing server to check rendered pillar data for ssh minions: https://github.com/moio/sumaform/pull/129

### DIFF
--- a/features/salt_formulas.feature
+++ b/features/salt_formulas.feature
@@ -33,6 +33,14 @@ Feature: Use salt formulas
      And I select "French (Canada)" in keyboard layout field
      And I click on "Save Formula"
      Then I should see a "Formula saved!" text
+     When I refresh the pillar data
+     Then the pillar data for "formulas" should be "- locale" on "sle-minion"
+     And the pillar data for "timezone:name" should be "Etc/GMT-5" on "sle-minion"
+     And the pillar data for "keyboard_and_language:keyboard_layout" should be "French (Canada)" on "sle-minion"
+     And the pillar data for "keyboard_and_language:language" should be "French" on "sle-minion"
+     And the pillar data for "formulas" should be empty on "ssh-minion"
+     And the pillar data for "timezone" should be empty on "ssh-minion"
+     And the pillar data for "keyboard_and_language" should be empty on "ssh-minion"
 
   Scenario: Apply the parametrized formula via the highstate
      Given I am on the Systems overview page of this "sle-minion"
@@ -46,9 +54,13 @@ Feature: Use salt formulas
 
   Scenario: Check the pillar data
      When I refresh the pillar data
-     Then the pillar data for "timezone:name" should be "Etc/GMT-5"
-     And the pillar data for "keyboard_and_language:keyboard_layout" should be "French (Canada)"
-     And the pillar data for "keyboard_and_language:language" should be "French"
+     Then the pillar data for "formulas" should be "- locale" on "sle-minion"
+     And the pillar data for "timezone:name" should be "Etc/GMT-5" on "sle-minion"
+     And the pillar data for "keyboard_and_language:keyboard_layout" should be "French (Canada)" on "sle-minion"
+     And the pillar data for "keyboard_and_language:language" should be "French" on "sle-minion"
+     And the pillar data for "formulas" should be empty on "ssh-minion"
+     And the pillar data for "timezone" should be empty on "ssh-minion"
+     And the pillar data for "keyboard_and_language" should be empty on "ssh-minion"
 
   Scenario: Reset the formula on the minion
      Given I am on the Systems overview page of this "sle-minion"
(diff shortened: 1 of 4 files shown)


features/step_definitions/navigation_steps.rb
```diff
@@ -230,6 +230,13 @@
   step %(I follow "#{$client_fullhostname}") if target == "sle-client"
 end
 
+When(/^I check the "(.*?)" client$/) do |target|
```
MalloZup: @meaksh  use here this function, 

https://github.com/SUSE/spacewalk-testsuite-base/blob/d3392f6518166c10aafdb4560b6861c268d4cecb/features/step_definitions/command_steps.rb#L14

```node.run("hostname -f")``` 
we should add to the object node.hostname, but imho is better then the one here

File: features/step_definitions/salt_steps.rb [hunk 2]
```diff
@@ -410,6 +410,17 @@
   end
 end
 
+When(/^I uncheck the "([^"]*)" formula$/) do |formula|
+  # Complicated code because the checkbox is not a <input type=checkbox> but an <i>
```
MalloZup: LOL . should be open a bug for that?

features/step_definitions/salt_steps.rb
```diff
@@ -474,11 +495,42 @@
...
+  elsif minion == "ceos-minion"
+    target = $ceos_minion_ip
+    cmd = "salt-ssh"
+    extra_cmd = "-i --roster-file=/tmp/tmp_roster_tests"
+    $server.run("printf '#{target}:\n  host: #{target}' > /tmp/tmp_roster_tests")
+  else
+    fail "Invalid target"
+  end
+  output, _code = $server.run("#{cmd} '#{target}' pillar.get '#{key}' #{extra_cmd}")
+  print output
```
MalloZup: why prints and not puts?, which is better for \n

https://stackoverflow.com/questions/5018633/what-is-the-difference-between-print-and-puts

features/step_definitions/salt_steps.rb
```diff
@@ -474,11 +495,42 @@
...
   fail unless output.split("\n")[1].strip == value
 end
 
+Then(/^the pillar data for "([^"]*)" should be empty on "([^"]*)"$/) do |key, minion|
+  if minion == "sle-minion"
+    target = $minion_ip
```
MalloZup: as a side remark, we have now also ssh-minion, do we test that for this target?

(centos7 minion, is a ssh or service)

features/salt_formulas.feature
```diff
@@ -57,6 +69,14 @@ Feature: Use salt formulas
      And I click on "Clear values" and confirm
      And I click on "Save Formula"
      Then I should see a "Formula saved!" text
+     And I refresh the pillar data
+     And the pillar data for "formulas" should be "- locale" on "sle-minion"
+     And the pillar data for "timezone:name" should be "CET" on "sle-minion"
+     And the pillar data for "keyboard_and_language:keyboard_layout" should be "English (US)" on "sle-minion"
+     And the pillar data for "keyboard_and_language:language" should be "English (US)" on "sle-minion"
+     And the pillar data for "formulas" should be empty on "ceos-minion"
+     And the pillar data for "timezone" should be empty on "ceos-minion"
```
MalloZup: @meaksh ,

looking at your test here:

https://github.com/SUSE/spacewalk-testsuite-base/pull/295/files#diff-8aef6aa74851e0d058f469ae0e69142fR130

you run commands when the ceos-minion is not bootstrapped.

https://github.com/SUSE/spacewalk-testsuite-base/blob/master/run_sets/testsuite.yml#L71

Imho, we need a new feature only for that, i would put it at the end.

meaksh: Thanks for your feedback @MalloZup ! 👍 

Would you be ok if we move the `salt_formula.feature` to be executed after `mincentos7` is bootstrapped? That way we would prevent from repeating all the formula stuff to test leaking on that new `salt_formula_leaks.feature`.

I would move `salt_formula.feature` after bootstraping centos7, as we can consider it now as dependent of both minions (sle-minion and ceos-minion).

What do you think?

meaksh: Well, forget last comment 😄 . I'll try using `ssh-minion` 👍

features/step_definitions/salt_steps.rb
Bischoff: There was one, if you remember.

features/step_definitions/salt_steps.rb
Bischoff: Is this print needed at all? It can be useful for debugging, but should it stay in the final testsuite?

features/step_definitions/salt_steps.rb
Bischoff: AFAIR we don't

features/step_definitions/salt_steps.rb
MalloZup: it doesn't harm a puts or a print, yo can see it even on debug so it's ok

features/step_definitions/salt_steps.rb
MalloZup: a bug on bugzilla not on github

meaksh:
I've updated the PR to use `ssh-minion` and address received feedback.

@Bischoff - I need to cross-check pillar data between `sle-minion` and `ssh-minion` in the same scenario because I want to ensure no leak is produced between those minions when a formula is assigned to `sle-minion`.

IMHO on this particular case, both minions should be tested on the same scenario. What do you think?

I'm about to run the testsuite on jenkins using this branch as @MalloZup suggested. I'll comment here back with the results.

BTW, enjoy your vacations @MalloZup ! 😄

Bischoff:
Ah OK, if it is intentional and motivated, then okay!

Thanks for the explanation.

meaksh:
Seems `manager-TEST-cucumber` is not working as expected.

This has been tested locally and it's working. I'm merging it now.

PR #296, opened 2017-06-19
Feature "minion_datacenter" broken if IPv6 not working, disabling for now

Bischoff:
Test fails with

`http://centos.s.uw.edu/centos/6/os/x86_64/Packages/gpm-libs-1.20.6-12.el6.x86_64.rpm:
[Errno 14] PYCURL ERROR 7 - "Failed to connect to 2607:4000:200:4b::120: Network is unreachable"`

Disabling it for now. Long term solution is either to add "ip_resolve=4" to /etc/yum/yum.conf, or (better) to make sure that IPv6 works on our test machines.

### DIFF
--- a/run_sets/testsuite.yml
+++ b/run_sets/testsuite.yml
@@ -91,8 +91,10 @@
 
 ### Scalability tests (after debug always)
 - features/docker_xmlrpc.feature
-- features/minion_datacenter.feature
-## smdba and change password feature always put them at the end of the testsuite.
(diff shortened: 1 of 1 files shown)


MalloZup:
if only the rhel6 is broken because of ipv6,

just remove here the rhel6  on this array
https://github.com/SUSE/spacewalk-testsuite-base/blob/d3392f6518166c10aafdb4560b6861c268d4cecb/features/step_definitions/command_steps.rb#L325

PR #299, opened 2017-06-28
Accept repository names to be surrounded by whitespace inside <a>

Bischoff:
Step ```When I select the "Test-Repository-x86_64" repo``` was failing because of new indentation.

```
      <a href="/rhn/channels/manage/repos/RepoEdit.do?id=500">
            Test-Repository-x86_64
      </a>
```
was not recognized as ```Test-Repository-x86_64```, because of the whitespace surrounding it.

The new XPath code uses ```normalize-space()``` instead of ```text()```. It is not a perfect solution, because the leading and trailing spaces are not only trimmed, but spaces inside the content are also collapsed. As a consequence, after this change, it might be needed to normalize spaces inside the names of repositories.

### DIFF
--- a/features/create_repository.feature
+++ b/features/create_repository.feature
@@ -36,7 +36,7 @@ Feature: Adding repository to a channel
     And I follow "Overview" in the left menu
     And I follow "Test-Channel-x86_64"
     And I follow "Repositories" in the content area
-    When I select the "Test-Repository-x86_64" repo
+    When I select the "Test-Repository-x86_64" repo and delete all whitespaces
     And I click on "Update Repositories"
     Then I should see a "Test-Channel-x86_64 repository information was successfully updated" text
 
@@ -65,7 +65,7 @@ Feature: Adding repository to a channel
     And I follow "Overview" in the left menu
     And I follow "Test-Channel-i586"
     And I follow "Repositories" in the content area
-    When I select the "Test-Repository-i586" repo
+    When I select the "Test-Repository-i586" repo and delete all whitespaces
     And I click on "Update Repositories"
     Then I should see a "Test-Channel-i586 repository information was successfully updated" text
 
--- a/features/step_definitions/content_steps.rb
+++ b/features/step_definitions/content_steps.rb
@@ -183,7 +183,8 @@ Then(/^I select the "([^"]*)" repo $/) do |arg1|
     first('input[type=checkbox]').set(true)
   end
 end
-
+# HACK: this step while cancel all whitespaces and also the steps
+# inside the repo name
 Then(/^I select the "([^"]*)" repo and delete all whitespaces$/) do |arg1|
   within(:xpath, "//a[normalize-space()='#{arg1}']/../..") do
     first('input[type=checkbox]').set(true)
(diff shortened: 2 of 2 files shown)


Bischoff:
Aaah the capybara way... I did not find about it!!!

Yes, "text" field in capybara is always trimmed from outer spaces. That should work, even without regexes.

We have an emergency here, so if I can't get it to work in a quick fashion I would accept this pull request "as is" as a temporary measure.

MalloZup:
@Bischoff , i would rename the step before accepting the PR. 
Just rename the step like " i select the "XX" repo and delete all whitespace"

This cost you 2 minutes, then merge the pr

MalloZup:
@Bischoff , ok i did for you, this time but remember that even on emergencies we need to take care in future and we should protect the us against this fast fixes.
Putting notes, and comments, because  what we fixed today with HACKs we will forget tomorrow. 

So even if i liked the PR, saying that there is an emergency whatever, doesn't justify the merge immediate of a PR without 2/3 min precautions. ( i can understand if the fix require 400 hours)

Bischoff:
your pull request is incorrect !

why didn't you wait ? I have a working solution based on your idea

now I must revert that

Bischoff:
dmaiocchi, "When I select the "Test-Repository-x86_64" repo and delete all whitespaces" is FALSE, we don't remove any whitespaces, we ignore them in the test.

Step names should reflect what a step does.

Bischoff:
starting to revert all that pull request, then applying a correct one :-(

MalloZup:
@Bischoff  ok sorry, then just commit directly without reverting, just change the name.

If i act like that, because you were menacing to merge it directly

"
We have an emergency here, so if I can't get it to work in a quick fashion I would accept this pull request "as is" as a temporary measure.
"

MalloZup:
you don't need to revert, just apply a commit, where you change the name better.

Bischoff:
this pull request i did, not something completely wrong

MalloZup:
@Bischoff , i didn't do something completely wrong, please don't use absolutism 👯‍♂️ 

What we can learn:
- don't put on PR "i have an emergency, i will fix it, or merge"

the rest is assume

Bischoff:
and the name you give is WRONG

Cucumber should be clear English description of what it does, not detail implementation. And these details were wrong too, we don't remove any whitespace, we normalized whitespace during the test

And NO I did not ask you to merge with changes i did not approve

MalloZup:
@Bischoff  i know cucumber, it's ok. what you say and so on, not everybody is english native etc
@Bischoff next time, just don't menage to merge if this is not ok for you, the rest i will not merge it then in future, ciao 😸

Bischoff:
I said:

>    We have an emergency here, so if I can't get it to work in a quick fashion I would accept this pull request "as is" as a temporary measure.


It means:

> I want to try your idea, but if I don't succeed in a short time, the I would accept my own pull request

Sorry if that was not clear.

I redid a new pull request and now everything is fixed. By the way your idea of using capybara's finders instead of xpath (and more precisely :text attribute) works nicely :smile: .

PR #302, opened 2017-07-01
Leave time for minions to appear in the preview list

Bischoff:
This is an experimental change; unfortunately, I need to test it in real
context, as I can't reproduce the problem in isolation. It will be revertied
if useless.

### DIFF
--- a/features/salt_ssh.feature
+++ b/features/salt_ssh.feature
@@ -48,9 +48,9 @@ Feature: 1) Bootstrap a new salt host via salt-ssh
     And I should see a "Remote Commands" text
     Then I enter command "rpm -q salt-minion"
     And I click on preview
-    And I wait for "3" seconds
+    And I wait for "10" seconds
     And I click on run
-    And I wait for "3" seconds
+    And I wait for "10" seconds
     And I expand the results for "ssh-minion"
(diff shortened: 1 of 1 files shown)


MalloZup:
@Bischoff, in head and 31
we have like a dinamic wait, wait until i see run, embedded on the function I click on run, so maybe you can backport it to 30 if wish . Otherwise i am also fine for this

MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base/blob/c786f91be7bc8936f697a9b52e206c98f4463517/features/step_definitions/salt_steps.rb#L306

Bischoff:
Ah, I did not think at comparing with head and 3.1. Right, a dynamic wait is better.

I will try to backport the dynamic wait to 3.0 if I find two seconds.

I am also curious to see if this stone age fix works. If think it will, according to one of the runs this morning, that made clear the problem was indeed ssh minions need more time to appear in the preview list.

MalloZup:
In head it was removed  because it was unstable, and for gmc etc, was not a green test and so on.
I think the backport is not difficult, and we will maintain at least 30 for longt time, at least tbe stability if we can make it better on reducinf false failure.

PR #303, opened 2017-07-03
Restoring a unified maximum timeout

Bischoff:
Set to a generous 10 minutes
The tests are suppose to return much earlier in case of success


MalloZup:
ok ![pap] ( https://media.giphy.com/media/hTh9bSbUPWMWk/giphy.gif)

Issue #301, opened 2017-06-30
REMOVE SMBDA from core-features


MalloZup:

Bischoff:
@dmaiocchi, please define "core-features" -- I am not sure to understand what this issue is about

MalloZup:
@Bischoff https://github.com/SUSE/spacewalk-testsuite-base/blob/master/run_sets/core-feature.yml#L9

bofh say to me that is not core, so we remove it from the list.

MalloZup:
fixed and adressed

PR #305, opened 2017-07-04
Make testsuite idempotent

MalloZup:
# What is this?

This branch represent the working copy of the idempotent testsuite.

# What does this mean?

We want to create a testsuite in which every test can be re-run without resetting anything.

# Development

- [x] create a working list of core-features that will init the server ( the core-list is not supposed at this moment to be idempotent)

https://github.com/SUSE/spacewalk-testsuite-base/blob/head-idempotent/run_sets/testsuite.yml

- [x] create a secondary list, of features that can be runned in random order and are **idempotent**

- [ ] everything should work on jenkins job test

**Note**: Jenkins Job is not using "centos7" and minionssh machines, due to lack of VMs.

## How to contribute

**Run locally your changes** and then open a PR *against `head-idempotent`*

# Infra

This branch will be tested against CI on:
https://ci.suse.de/view/Manager/view/Manager-Test/job/manager-TEST-cucumber/

Full Terraform setup:
https://gitlab.suse.de/galaxy/sumaform-test-runner/blob/master/test/main-full.tf


MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base/issues/301

run_sets/core.yml
```diff
@@ -0,0 +1,12 @@
+# These features must be runned to initialize the server
+- features/setup_user_and_proxy.feature
+- features/mainpage.feature
+# smdba before we start to sync channels
+- features/smdba.feature
```
MalloZup: smbda is not core

If we take this approach with multiples file.yml, we need to consider:
- [ ] modify the rakefile (output.html) is hardocded and will be overwritten by each .yaml (create dyn one)
- [ ] we will end up with having 2 output.html or more, which for reviewing is a big change, ( in past we have only 1 file.html)

MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base/pull/322

PR #307, opened 2017-07-04
Master remove hardcoded names of clients

Bischoff:
... and generalize the use of DEFAULT_TIMEOUT

(WIP: should work, but not tested yet)

### DIFF
--- a/features/step_definitions/navigation_steps.rb
+++ b/features/step_definitions/navigation_steps.rb
@@ -8,7 +8,7 @@ Then(/^the current path is "([^"]*)"$/) do |arg1|
   fail unless (current_path == arg1)
 end
 
-And(/^I wait until i see "([^"]*)" text$/) do |text|
+When(/^I wait until i see "([^"]*)" text$/) do |text|
   begin
     Timeout.timeout(150) do
       loop do
@@ -21,7 +21,7 @@ And(/^I wait until i see "([^"]*)" text$/) do |text|
   end
 end
 
-And(/^I wait until i see "([^"]*)" text, refreshing the page$/) do |text|
+When(/^I wait until i see "([^"]*)" text, refreshing the page$/) do |text|
   begin
     Timeout.timeout(150) do
       loop do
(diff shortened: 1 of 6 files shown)


mantel:
Looks very good to me!

features/step_definitions/salt_steps.rb
```diff
@@ -756,8 +756,9 @@
 
 Given(/^the salt-master can reach "([^"]*)"$/) do |minion|
   begin
+    # 300 is the default 1st keepalive interval for the minion
     # where it realizes the connection is stuck
-    Timeout.timeout(DEFAULT_TIMEOUT) do
+    Timeout.timeout(300) do
```
MalloZup: i would use here like 

```
  KEEP_ALIVE_TIMEOUT= 300
  Timeout.timeout(KEEP_ALIVE_TIMEOUT) do
```

features/step_definitions/salt_steps.rb
```diff
@@ -25,7 +25,7 @@
     start = Time.now
     # 300 is the default 1st keepalive interval for the minion
     # where it realizes the connection is stuck
-    Timeout.timeout(DEFAULT_TIMEOUT) do
+    Timeout.timeout(300) do
```
MalloZup: Can you introduce a var here?

Otherwise it seems a Magic number

MalloZup: a var like KEEP_ALIVE_TIMEOUT

features/step_definitions/navigation_steps.rb
```diff
@@ -34,6 +34,24 @@
     raise "Couldn't find the #{text} in webpage"
   end
 end
+
+When(/^I wait until I see the name of "([^"]*)", refreshing the page$/) do |target|
+  case target
```
MalloZup: here,  we see that we should get rid of the var $minion_fullhostname and make it as an attribute of the object.

THen we could use 
```get_node(target) ```
```  step %(I wait until i see "#{node.fullhostname}" text, refreshing the page```

Bischoff:
Very nice idea, the variable (from a ruby pov that's a constant, because it's upper case)

We already have this comment (don't know who did it), but having the variable too does not harm.

Bischoff:
Ouchie, it looks like i'm missing a call to get_node, nice catch dario

Bischoff:
> here, we see that we should get rid of the var $minion_fullhostname and make it as an attribute of the object.

I examined this - you're right, but it's a big change that affect many places. So I will not do it now. We need to put it on the refactoring TODO and do it correctly in a different pull request.

MalloZup:
@Bischoff sure, that's why i opened an issue, because for this PR is to much.

PR #308, opened 2017-07-04
[WIP]Modify core-features

MalloZup:


run_sets/testsuite.yml
```diff
@@ -4,28 +4,28 @@
...
 # SECONDARY FEATURE:
 
 # this feature can be runned in random order.
 # FIXME: MAKE THIS HAPPEN
 
-
-
+# server secondary 
+- features/mainpage.feature
+- features/create_activationkey.feature
```
MalloZup: since this feature, create_activationkey create key that are reused, could be considered as core feature, imho is not because we can make feature create theirs act-key

MalloZup:
close, use the other branch

PR #313, opened 2017-07-05
Prevent spurious failures for presence ping test

meaksh:
This should prevent the spurious failures caused in case an state is being executed when performing the presence ping test.

On such cases we don't expect to get the actual response from `state.show_highstate` but a response from the minion like: 

```
- The function "state.apply" is running as PID 8710 and was started at 2017, Jul 05 12:11:54.266463 with jid 20170705121154266463
```

Getting either any of both responses is a valid scenario as the minion is actually responding.

### DIFF
--- a/features/salt_software_states.feature
+++ b/features/salt_software_states.feature
@@ -84,7 +84,7 @@ Feature: Check the Salt package state UI
     Then I follow "States" in the content area
     And I follow "Highstate" in the content area
     And I wait for "6" seconds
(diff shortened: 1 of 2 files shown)


MalloZup:
@meaksh  backport the fix to 3'0 and 31 direcly

meaksh:
Merged on `master` and `manager-3.1`.

I noticed that the presence ping test is not present on `manager30` branch. I also backported the test to `manager30`.

PR #316, opened 2017-07-06
Fix: increase timeout when issuing cmd.run

mbologna:
It happens sometimes that we get a failure in these two scenarios,
because we are waiting for the command to reach all the minions.
A first workaround (this pr) increases the time we are waiting for the
button to appear.


MalloZup:
thx @mbologna

PR #317, opened 2017-07-06
Fix: update button text

mbologna:


mbologna:
Closing as no longer relevant (next run of testsuite will be run against updated packages with new strings)

PR #322, opened 2017-07-10
 Make systempage.feature to core-feature

MalloZup:


MalloZup:
@meaksh ,  i splitted the systempage feature , since there were clobber test.

We don't need cleanup scenario for systempage since is not adding/modifying stuff, it's just grepping for webui-content

features/distro_cobbler.feature
```diff
@@ -0,0 +1,157 @@
+# Copyright (c) 2010-2017 Novell, Inc.
+# Licensed under the terms of the MIT license.
+
+Feature: test cobbler and distro Kickstart
+
+  Background:
+  Given I am authorized
```
meaksh: Please, add two more spaces here for steps inside the `Background` block.

features/systemspage.feature
```diff
@@ -14,7 +14,6 @@ Feature: Explore the main landing page
 
   Scenario: Completeness of the side navigation bar and the content frame
     Then I should see a "System Overview" text
-    And I should see a "No systems." text
```
meaksh: Instead of removing the step, which is actually checking that a "No systems." message exist when there is no systems, why don't modify the step to make it idempotent?

For example, something like: 
```
And I should see a "No systems." or "mgr.suse.de" text
```

That way, we're not loosing power but increasing it! 😄 

What do you think?

features/distro_cobbler.feature
MalloZup: ok thx

features/systemspage.feature
MalloZup: 🕺 
with "mgr.suse.de" we are hardoding the system. We need to keep care that the testsuite will run on a cloud soon, can be run in a local machine. So the mgr.suse.de.

Imho, since the test is idempotent, we cannot expect such a test. And i was thinking, but i see no advantage to grep that "there should be a 0 systems". 

It was not a good test 🚒

run_sets/testsuite.yml
```diff
@@ -8,6 +8,7 @@
 - features/channels_add.feature
 - features/push_package.feature
 - features/create_repository.feature
+- features/systemspage.feature
```
MalloZup: looking at the original https://github.com/SUSE/spacewalk-testsuite-base/blob/master/run_sets/testsuite.yml#L9

MalloZup:
@meaksh  well at the beginning it was everything a core-feature 💠

PR #324, opened 2017-07-10
Using random salt-thin dir for each salt-ssh call

meaksh:
This avoid a race condition when multiple salt-ssh calls are trying to deploy salt-thin using the same salt-thin dir. The "-w" parameter would also wipe the deployed salt-thin after each salt-ssh execution.

This will fix the spurious failure of "# Scenario: Parametrize the formula on the minion".

This would go also to `head` branch.


MalloZup:
💘

meaksh:
Merged on `manager31` and `master`

PR #326, opened 2017-07-11
Make 'salt.feature' idempotent

meaksh:
Also this PR updates the list of salt idempotent tests.

### DIFF
--- a/features/salt.feature
+++ b/features/salt.feature
@@ -22,7 +22,7 @@ Feature: Salt is configured and running
     Then salt-minion should be running on "sle-minion"
 
   Scenario: The "sle-minion" appears in the master's incoming queue
-    Then the list of the "pending" keys should contain "sle-minion" hostname
+    Then the list of the "all" keys should contain "sle-minion" hostname
 
    Scenario: There are no top.sls file in certain folders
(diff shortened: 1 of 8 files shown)


features/salt.feature
```diff
@@ -22,7 +22,6 @@ Feature: Salt is configured and running
     Then the Salt Minion should be running
 
   Scenario: The minion appears in the master's incoming queue
-    When I list unaccepted keys at Salt Master
     Then the list of the keys should contain this client's hostname
```
MalloZup: @meaksh  i would just change this "then" with

``Then the list of the keys should contain this  sle-minion hostname``

sle-minion is normal salt
ssh-minion is sles-minion ssh managed

MalloZup: we could even parametrize this step, but afaik we don't use for ssh-minions

MalloZup:
## refractor: 
- [ ] remove this : https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/salt_minion_details.feature#L23

we have a generic steps

MalloZup:
nice work

MalloZup:
[go go Jenkins! ] 💃

PR #327, opened 2017-07-11
Small fix

MalloZup:


MalloZup:
@meaksh  if nothing against, we need to merge this to fix the head-idempotent branch

features/salt_minions_page.feature
```diff
@@ -6,9 +6,6 @@ Feature: Explore the Minions page
   As an authorized user
   I want to verify all the minion key management features in the UI
 
-  Background:
-    Given I am authorized as "testing" with password "testing"
-
   Scenario: Delete sles-minion system profile (Explore the Minions page)
     Given I am on the Systems overview page of this "sle-minion"
```
meaksh: Does this scenario need also the "Given I am authorized as "testing" with password "testing" step? Like all other scenarios?

MalloZup: no, it doesnt

MalloZup: Imho, i'm not  a big fun of background, only if it spare really lot of steps. 
otherwise it make less clear a single scenario

MalloZup:
thx @meaksh , merging, let's make it green. ( core-features are growing, but they are necessary) salt-ssh etc

PR #330, opened 2017-07-12
Make some features idempotent

MalloZup:
- [x] custom_info


MalloZup:
@meaksh green is the future

PR #336, opened 2017-07-13
Idempotent3

MalloZup:


features/check_registration.feature
```diff
@@ -17,7 +17,6 @@ Feature: Check client registration
     And I should see a "Packages:" link
     And I should see a "Edit These Properties" link
     And I should see a "[Management]" text
-    And I should see a "Add to SSM" link
```
MalloZup: here was causing pb

meaksh: @MalloZup could you please revert this change from PR? I'll address the needed changes of `check_registration.feature` on my upcoming PR. That way we can merge this before even my PR.

features/action_chain.feature
```diff
@@ -58,20 +58,6 @@ Feature: Test action chaining
     And I click on "Confirm"
     Then I should see a "Action has been successfully added to the Action Chain" text
 
-  Scenario: I add a config file deployment to the action chain
```
meaksh: Why not create a cleanup scenario to remove the deployed files instead of removing the entire scenario?

features/delete_system_from_conf.feature
```diff
@@ -1,15 +0,0 @@
-# Copyright (c) 2015 SUSE LLC
-# Licensed under the terms of the MIT license.
-
-Feature: Remove system from configuration channel
-
-  Scenario: Remove system from configuration channel
```
meaksh: Why not using this scenario as cleanup scenario for `Scenario: I add a config file deployment to the action chain` + some steps to remove the deployed files on the "sle-client" ?

features/action_chain.feature
MalloZup: 🕶  We need to create a config channel for that, and  assign the files to the config-channel
Then the cleanup. Since the feature is to test action-chain, i was thinking that we could get rid of that.

features/delete_system_from_conf.feature
MalloZup: ok i have added this scenario on latest cmit

MalloZup:
ok 🐼

PR #339, opened 2017-07-13
Move spacewalk_repo_sync.feature to idempotent

MalloZup:
Feature is idempotent since create base channel is on core-feature


MalloZup:
@meaksh  yes we did it !
po

PR #341, opened 2017-07-13
head-idempotent - wait default time for images

Bischoff:
We were waiting only 20 x 20 = 400 seconds for images to build, that's not enough on slow machines

Now using standard DEFAULT_TIMEOUT, with a smaller sleep

NOT TESTED YET


MalloZup:
@Bischoff  ok thx, i will go to merge this PR, we can see it on next run if it works.

PR #355, opened 2017-07-19
Modify smdba test to work with postgresql94 as well as with postgresql96

mantel:
With new postgresql version, tablespace "postgres" seems to have been renamed to "template1". So modify test to work on both versions (check for template0 and susemanager).


MalloZup:
@mantel thx! as discussed, make a pr against head-idempotent and 30.

The head-idempotent will be merged on head and 31

PR #359, opened 2017-07-19
Make action_chain idempotent

MalloZup:
Rename it with right prefixes


MalloZup:
@meaksh i will merge this, so we will have tomorrow less failures. I made idempotent feature and renamed the feature according the prefix, plus some doc

PR #360, opened 2017-07-19
make smdba config check work for postgresql96 too

mantel:
Tested against postgresql96 only so far.


mantel:
Also tested with postgresql94. Ready to merge.

PR #362, opened 2017-07-20
fix core_trad_client

MalloZup:


features/core_trad_register_client.feature
```diff
@@ -26,14 +26,10 @@ Feature: Register a client
     And I should see a "Delete System" link
     And I should see a "Initial Registration Parameters:" text
     And I should see a "OS: sles-release" text
-    And I should see a "Release: 12.2" text
```
MalloZup: more generic ( this is also tested in allcli_overview)

features/step_definitions/content_steps.rb
```diff
@@ -23,6 +23,17 @@
     fail unless page.has_content?(arg1)
   end
 end
+
+#
+# Test for a text in the whole page
+#
```
MalloZup: @meaksh  you implemented already a similar step afaik?

features/step_definitions/content_steps.rb
```diff
@@ -23,6 +23,17 @@
...
+Then(/^I should see a "([^"]*)" or "([^"]*)" text$/) do |text, text2|
+  unless page.has_content?(text) || page.has_content?(text2)
+    sleep 25
```
meaksh: I would remove this sleep from here, as we don't want to wait in case we are reusing this step in other feature. I'd put a "wait step" in the above scenario 😄

features/core_trad_register_client.feature
meaksh: Great!! 👍

features/core_trad_register_client.feature
```diff
@@ -16,7 +16,7 @@ Feature: Register a client
     And I click on the css "a#clear-ssm"
     And I wait for the data update
     Then I should see a "System Status" text
-    And I should see a "Software Updates Available" text
+    And I should see a "Software Updates Available" or "System is up to date" text
```
meaksh: Do we have another feature or test where we check if `Software Updates Available` label is shown when updates are available?

If this is the only place where we check it, then I would not change this step and would try wait until the errata information is refreshed or some updates are actually available. 

What do yo think?

features/step_definitions/content_steps.rb
MalloZup: good catch :tada: :tiger:

MalloZup: @meaksh  stuff there is when maybe we need to wait for a javascript code to be executed, we can make it to 10 sec imho

features/core_trad_register_client.feature
MalloZup: in future, we will might have upd2date systems or not

meaksh: I understand that, but as this is a core feature, we should decide whether it should show one or other label depending on the actual status of the traditional client at this point.

Also notice that vendor channels (created by core_srv_sync_channels) are created and assigned after this step, in the latest core feature step. So it's not expected software updates unless you provide them using some custom repository (like virgo, andromeda, etc).

MalloZup:
@meaksh  removed and renamed some feature with conventions

MalloZup:
:japanese_ogre:

PR #369, opened 2017-07-21
Remove script in tools

MalloZup:
I readded the people in CC. 
The discussion is better in github.

Why we need this removal:
1) Initially , the script never passed  a PR was pushed directly
2) Since we have a  "core feature" set, you don't need to setup your server for minimal dev. anymore 
3) We want only 1 official way to run the testsuite. (custom script should remain on custom repo)
4)The script is not a safe option because:
   - hardcoded:  if the scenario name change, it will fail ( Maintenance cost). ( if we remove scenario etc) 
   - differ-from features
    -more serious: this script allow to deviate from core-features setup, which imply that we are not 100% follow the upstream, so the behaviour becomes unpredictable. 

I would suggest to keep the script on your custom repo, and we can still have a custom note in doc, to say how to run scenarios.

@Bischoff , i will keep this pr open until you are back from holiday, so  you can comment once you are back

@meaksh , @mantel @moio @mcalmer  need your pov.


mantel:
Go ahead. The script was meant to be only temporary from the beginning anyway.

MalloZup:
@Bischoff ^^

PR #374, opened 2017-07-27
Head idempotent: Improve step to check taskomatic job status

meaksh:


meaksh:
I'm testing this on ctl node but think I missed something, let's keep this as WIP until get fixed.

PR #385, opened 2017-08-07
autoimport gpgkey from channel for ssh-minions

MalloZup:
:panda_face:


MalloZup:
i think this will fix the issue, i will just rerun it, if openvpn stay stable

MalloZup:
@hustodemon @dincamihai  do we need to make zypper ref gpg... command after importing the key via sumaform? or this is not needed more

PR #387, opened 2017-08-07
Click on CSS button to refresh the page instead manual page refresh

meaksh:
This PR fixes an issue caused due a waiting step which refreshes the page in a wrong way (without keeping the package list filter). 🍰  

These changes will go to `manager31` and `master` branches.

### DIFF
--- a/features/step_definitions/salt_steps.rb
+++ b/features/step_definitions/salt_steps.rb
@@ -449,7 +449,7 @@ end
 Then(/^I click on the css "(.*)" until page does not contain "([^"]*)" text$/) do |css, text|
   not_found = false
   begin
-    Timeout.timeout(30) do
+    Timeout.timeout(DEFAULT_TIMEOUT) do
       loop do
         unless page.has_content?(text)
           not_found = true
@@ -467,7 +467,7 @@ end
 Then(/^I click on the css "(.*)" until page does contain "([^"]*)" text$/) do |css, text|
   found = false
   begin
-    Timeout.timeout(30) do
+    Timeout.timeout(DEFAULT_TIMEOUT) do
       loop do
         unless page.has_content?(text)
           found = true
--- a/features/step_definitions/salt_steps.rb
+++ b/features/step_definitions/salt_steps.rb
@@ -446,12 +446,12 @@ Then(/^I manually remove the "([^"]*)" package in the minion$/) do |package|
   $minion.run(cmd, false)
 end
 
-Then(/^I click on the css "(.*)" until page does not contain "([^"]*)" text$/) do |css, arg1|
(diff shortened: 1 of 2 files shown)


features/step_definitions/salt_steps.rb
```diff
@@ -459,10 +459,29 @@
       end
     end
   rescue Timeout::Error
-    raise "'#{arg1}' cannot be found after several tries"
+    raise "'#{arg1}' is still found after several tries"
   end
   raise unless not_found
 end
+
+Then(/^I click on the css "(.*)" until page does contain "([^"]*)" text$/) do |css, arg1|
```
MalloZup: can we put "text" instead of arg1?

meaksh: Yes, we can! 👍 😄

features/step_definitions/salt_steps.rb
```diff
@@ -459,10 +459,29 @@
...
+  found = false
+  begin
+    Timeout.timeout(30) do
```
MalloZup: we could use DEFAULT_TIMEOUT here

MalloZup: or we could make MIN_DEFAULT_TIMEOUT label instead of magic numbers..

MalloZup:
robo

features/step_definitions/salt_steps.rb
```diff
@@ -446,23 +446,42 @@
...
+        unless page.has_content?(text)
           not_found = true
           break
         end
         find(css).click
       end
     end
   rescue Timeout::Error
-    raise "'#{arg1}' cannot be found after several tries"
+    raise "'#{text}' still found after several tries"
```
ncounter: I think the meaning should be "not yet found`, am I getting it wrong?

meaksh: Not in this case. This particular step is actually failing if `#{text}` is still found after the timeout is reached.

The case you're mentioning is just covered by the next step implemented on this file: https://github.com/SUSE/spacewalk-testsuite-base/pull/387/files#diff-daf5562d26751e108a56ac4ad5257bc9R480

ncounter: Aaaahh, I see! Thank you for explaining it. 👍

MalloZup:
@meaksh  we need a backport for head too.

- [ ] backport to head

meaksh:
Merged on `manager31` and `master`.

MalloZup:
ok thx @meaksh :rose:

Bischoff:
Yes, please use the default timeout. It is very long (10 minutes), but as long as you can exit the loop earlier, that's fine, it will be so long only in case of an unrecoverable failure.

PR #391, opened 2017-08-08
Generate bootstrap-repo os_version based dynamically

ncounter:
Generate the `bootstrap-repo` based on the `os_version` dynamically reading the `os_version` directly from the client.

### DIFF
--- a/features/core_min_bootstrap.feature
+++ b/features/core_min_bootstrap.feature
@@ -3,9 +3,9 @@
 
 Feature: register a salt-minion via bootstrap
 
-  Scenario: Create bootstrap-repo for sle12sp2
+  Scenario: Create bootstrap-repo for a salt client
      Given I am authorized
-     And  I run "mgr-create-bootstrap-repo -c SLE-12-SP2-x86_64" on "server" without error control
+     And I create the "x86_64" bootstrap-repo for "sle-minion" on the server
 
   Scenario: bootstrap a sles minion (it will be deleted after)
      Given I am authorized
--- a/features/step_definitions/command_steps.rb
+++ b/features/step_definitions/command_steps.rb
@@ -395,3 +395,13 @@ And(/^I wait until the package "(.*?)" has been cached on this "(.*?)"$/) do |pk
     end
   end
 end
+
+And(/^I create the "([^"]*)" bootstrap-repo for "([^"]*)" on the server$/) do |arch, target|
+  node = get_target(target)
+  os_version_raw, _code = node.run('grep "VERSION=" /etc/os-release')
+  os_version = os_version_raw.strip.split('=')[1].delete '"'
+  puts 'i should see the os version: ' + os_version
+  command = 'mgr-create-bootstrap-repo -c SLE-' + os_version + '-' + arch
+  puts 'running the command on the server:' + command
+  $server.run(command, false)
+end


features/step_definitions/bootstrap_steps.rb
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2010-2017 SUSE-LINUX
```
MalloZup: we can put this step in common.steps or command.steps.

We basically don't need to create a new step file for the bootstrap.

features/step_definitions/bootstrap_steps.rb
```diff
@@ -0,0 +1,12 @@
...
+# Licensed under the terms of the MIT license.
+
+And (/^I create the bootstrap-repo for "([^"]*)" on the "([^"]*)"$/) do |target, server|
+  node = get_target(target)
+  os_version_raw, _code = node.run('grep "VERSION=" /etc/os-release')
+  os_version = os_version_raw.strip.split('=')[1].delete '"'
+  puts 'i should see the os version: ' + os_version
+  command = "mgr-create-bootstrap-repo -c SLE-" + os_version + "-x86_64"
+  puts 'running the command on the server:' + command
+  _out, $fail_code = get_target(server).run(command, false)
```
MalloZup: here i would just use: 
```$server.run(command, false)```
since $server is a globall variable that we define in ```support`` dir. Every file that we have there, classes, function, vars, is automatically avaible for the feature and steps.

File: features/step_definitions/bootstrap_steps.rb [hunk 3]
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2010-2017 SUSE-LINUX
+# Licensed under the terms of the MIT license.
+
+And (/^I create the bootstrap-repo for "([^"]*)" on the "([^"]*)"$/) do |target, server|
```
MalloZup: here i would remove the server as a target
```And (/^I create the bootstrap-repo for "([^"]*)" on the server$/) do |target|```

because the $server.run , see last comment

File: features/core_min_bootstrap.feature [hunk 4]
```diff
@@ -3,9 +3,9 @@
 
 Feature: register a salt-minion via bootstrap
 
-  Scenario: Create bootstrap-repo for sle12sp2
+  Scenario: Create bootstrap-repo for a salt client
      Given I am authorized
-     And  I run "mgr-create-bootstrap-repo -c SLE-12-SP2-x86_64" on "server" without error control
+     And I create the bootstrap-repo for "sle-minion" on the "server"
```
MalloZup: @ncounter maybe we should call it 

``` And (/^I create the x64 bootstrap-repo for "([^"]*)" on the "([^"]*)"$/) do |target|```  

because we are assuming intel. but we could have also other arch maybe in future.

File: features/step_definitions/bootstrap_steps.rb [hunk 5]
```diff
@@ -0,0 +1,12 @@
...
+  node = get_target(target)
+  os_version_raw, _code = node.run('grep "VERSION=" /etc/os-release')
+  os_version = os_version_raw.strip.split('=')[1].delete '"'
+  puts 'i should see the os version: ' + os_version
+  command = 'mgr-create-bootstrap-repo -c SLE-' + os_version + '-x86_64'
+  puts 'running the command on the server:' + command
+  _out, $fail_code = get_target(server).run(command, false)
```
MalloZup: $server.run(command. false)

features/step_definitions/command_steps.rb
```diff
@@ -395,3 +395,13 @@
...
 end
+
+And (/^I create the "([^"]*)" bootstrap-repo for "([^"]*)" on the server$/) do |arch, target|
+  node = get_target(target)
+  os_version_raw, _code = node.run('grep "VERSION=" /etc/os-release')
+  os_version = os_version_raw.strip.split('=')[1].delete '"'
+  puts 'i should see the os version: ' + os_version
+  command = 'mgr-create-bootstrap-repo -c SLE-' + os_version + '-' + arch
+  puts 'running the command on the server:' + command
+  _out, $fail_code = $server.run(command, false)
```
MalloZup: nitpick>

$fail_code and out  is not needed here

```$server.run(command, false) ``` is enough

ncounter: Why not? The previous step this command was running was https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/step_definitions/command_steps.rb#L214? Maybe it was generic handling for other tests that use this step?

MalloZup: $fail_code is a used for some steps that need to check the fail_code, that"s why we use a global var for passing it to other steps

MalloZup:
gundam
space

@ncounter  you can backport this step to 3.1 branch.

PR #393, opened 2017-08-11
Dynamic way to enable repo

MalloZup:


MalloZup:
this will fix 4/5 failure actually on the 3.1 testsuite, and improve orthogonality with diff os-client sles

MalloZup:
@ncounter yop, fixed the englishhh

Issue #309, opened 2017-07-04
Create Hostname attribute for node object


MalloZup:
get_node(target)
step %(I wait until i see "#{node.fullhostname}" text, refreshing the page

MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base/pull/307/files#diff-e1dcc62083118a85850a986622970436R44

MalloZup:
will be fixed by pr

PR #396, opened 2017-08-17
integrate lavanda inside testsuite

MalloZup:
For future the mixin class is better to have it inside the testsuite


MalloZup:
fix https://github.com/SUSE/spacewalk-testsuite-base/issues/309

MalloZup:
- add a $server.hostname 
- add new method

PR #397, opened 2017-08-17
Idempotent test for bootstrapping system (salt-ssh + minion) via XMLR…

hustodemon:
…PC API


hustodemon:
Cherry-picked to manager31.

PR #404, opened 2017-08-23
[WIP]: 3.1: Update SSM features to match bsc1047702 fix

moio:
PR https://github.com/SUSE/spacewalk/pull/1919 changes the way SSM buttons work, this patch addresses consequent failures, listed here:

https://github.com/SUSE/spacewalk/pull/1919#issuecomment-321857106

PR #405, opened 2017-08-23
HEAD: Update SSM features to match bsc1047702 fix

moio:
Port of #404 to `master`


MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base/blob/4ba13c0c690a1e1d55d6f92b319fe790622d9ea0/features/srv_create_group.feature#L48

moio:
That should not be needed @MalloZup - AFAIU that test does not fail (and it is not expected to fail).

Merging now.

Issue #407, opened 2017-08-23
Scenario: Create bootstrap-repo is not working for sles11sp4


MalloZup:
The step 
```And I create the "x86_64" bootstrap-repo for "sle-minion" on the server``` is failing for sles11sp4 system

The problem is that:

we are executing the command
``` : mgr-create-bootstrap-repo -c SLE-11.4-x86_64```

this should be
```mgr-create-bootstrap-repo -c SLE-11-SP4-x86_64```

```console
NAME="SLES"
VERSION="12-SP2"
VERSION_ID="12.2"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP2"
``` 
```console
cat /etc/os-release 
NAME="SLES"
VERSION="11.4"
VERSION_ID="11.4"
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:11:4"

```

The same will be valuable for sles11sp3, os-release file between sles is not consistent

MalloZup:
@ncounter  i will fix it because i need it for sles11sp4

MalloZup:
fixed

PR #408, opened 2017-08-23
fix repo creation

MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base/issues/407


MalloZup:
@ncounter yop, i wads thinking that we could just hardcode and creating all the metadata for all repo.

Maybe for sles15 if os-release change , we could do it.
Thx :panda_face:

MalloZup:
thx @ncounter , the stuff was that we couldn"t imagine, that sles11sp4 was not CONSISTENT. :rofl:

PR #417, opened 2017-08-31
fix some sles11sp4 issues

MalloZup:


features/step_definitions/salt_steps.rb
```diff
@@ -104,17 +104,11 @@
...
-    target = $ceos_minion
-  else
-    raise 'no valid name of minion given! '
-  end
+  node = get_target(minion)
   i = 0
   MAX_ITER = 40
   loop do
-    _out, code = target.run('systemctl status salt-minion', false)
+    _out, code = target.run('rcsalt-minion status', false)
```
MalloZup: this was incopatible with sles11sp4, since was a  systemd free zone :dancer:

PR #419, opened 2017-09-01
Update VHM tests for the new page implementation

cbbayburt:
This PR updates Virtual Host Manager tests to comply with the changes implemented in [Manager-3.1: Kubernetes integration #1848](https://github.com/SUSE/spacewalk/pull/1848)

### DIFF
--- a/features/srv_virtual_host_manager.feature
+++ b/features/srv_virtual_host_manager.feature
@@ -12,22 +12,24 @@ Feature: Test "virtualhostmanager" Web UI.
   Scenario: Check VMware page
    Given I am on the Systems page
     When I follow "Virtual Host Managers"
-    And I follow "Add VMware-based Virtual Host Manager"
-    Then I should see a "Add a VMware-based Virtual Host Manager" text
-    And I should see a "Label:" text
-    And I should see a "Hostname:" text
-    And I should see a "Port:" text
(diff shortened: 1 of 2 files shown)


cbbayburt:
On the delete scenario, there are two buttons with the same name. The correct one has class `btn-danger` so we should additionally use a class selector.

delete-btn

cbbayburt:
It looks like Capybara cannot click on pop-up dialogs anyway. We'll need to update the method in `navigation_steps.rb`

MalloZup:
https://github.com/SUSE/spacewalk-testsuite-base/blob/4ba13c0c690a1e1d55d6f92b319fe790622d9ea0/features/srv_docker_advanced_test.feature#L13

cbbayburt:
@MalloZup i've tried with both `btn-danger` and `.btn-danger` but no luck
```
Unable to find button "btn-danger" (Capybara::ElementNotFound)
```

cbbayburt:
Ah, it should be an id, not a class

MalloZup:
@cbbayburt ```find("#btn-danger", visible: false).click```

cbbayburt:
But I need to update the DeleteDialog component first.

cbbayburt:
Merged and cherry-picked into `master`

PR #421, opened 2017-09-06
Handle modal dialog navigation

cbbayburt:
This PR adds some step definitions to navigate through modal dialogs.

### DIFF
--- a/features/srv_virtual_host_manager.feature
+++ b/features/srv_virtual_host_manager.feature
@@ -56,7 +56,8 @@ Feature: Test "virtualhostmanager" Web UI.
     When I follow "Virtual Host Managers"
     And I follow "file-vmware"
     And I click on "Delete"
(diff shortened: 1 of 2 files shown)


Bischoff:
Great xpath ! Congrats.

Much better gherkin, no internals exposed.

The "Wait" step does not wait, and it should  :smile:, otherwise we will continue to have this racing condition. You may take example on the step:

    /^I try to reload page until contains "([^"]*)" text$/

The trick is to do something along the lines:

```
  begin
     Timeout.timeout(DEFAULT_TIMEOUT) do
      loop do
        (blah)
      end
      sleep(1)
   end
   rescue Timeout::Error
      (blah)
   end
```

cbbayburt:
Thanks, @Bischoff, it should work now.

Bischoff:
Not tested, but looks good to me now.

PR #426, opened 2017-09-11
Sles11 family adaptions

MalloZup:
- fix rcsalt minion behaviour
- fix docker unavaible for sles11 (test were running there, we need to skip )


MalloZup:
@Bischoff  for the raise, we have already the check here:

https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/support/twopence_init.rb#L59

I removed the vars for cmd, i was thinking the line was to big for rubocop , but it"s ok

MalloZup:
@Bischoff  all your remarks adressed

PR #430, opened 2017-09-13
Remove test that test not effective data

MalloZup:
Additionaly the calls are tested in junit.

To mantain this hardcoded test doesn"t bring an effective value on the real testing, and we consume time and performance.


MalloZup:
as discussed, merging

PR #435, opened 2017-09-14
Manager31 fix modals

Bischoff:
This pull request *is incorect* however it enables to have a green testsuite.

The problem is that it would work even if the modal window has not been triggered at all. But that's the best we have so far.


Bischoff:
wrong branch

PR #441, opened 2017-09-19
Big testsuite cleanup: non-core features

Bischoff:
Second and last phase of big testsuite cleanup: applied conventions, fixed English.


MalloZup:
@Bischoff  for the travis red:
you have a duplicata scenario, and for rubocop since the gem (rubocop) got updated on the travis debian container, we have from time to time new failures.

This we need to fix them, but if this are out of scope of the PR, i would suggest you to run

``` console
cd spacewalk-testsuite-base
gem update rubocop
rubocop --auto-gen-config
git diff
```
So we can blacklist the new failures and fix them later. Of course this is only because the updated gem, we blacklist them :)

Bischoff:
Blacklisted travis problems as you suggested.

MalloZup:
thx

Bischoff:
thanks for the review and the merge

PR #443, opened 2017-09-21
Change all functions from camelCase to snake_case, including doc

juliogonzalez:
Tested locally.

Results at http://wotan.suse.de:8999/spacewalk-testsuite-base-head/output.html (two failed scenarios: one is a brittle test, the other one I would say it is as well)


MalloZup:
@juliogonzalez  feel free to cherry-pick it to 31.

PR #444, opened 2017-09-21
Change all functions from camelCase to snake_case, including doc

juliogonzalez:
This is a PR instead of cherry-pick because there were some minor conflicts.

Tested locally.

Results at http://wotan.suse.de:8999/spacewalk-testsuite-base-31/output.html (two failed scenarios: one is a brittle test, the other one I would say it is as well, as in the previous PR)


Bischoff:
lgtm, merging

PR #449, opened 2017-09-21
Apply conventions to cucumber 4

Bischoff:
next try


Bischoff:
rhaa wrong branches again

PR #450, opened 2017-09-22
Normalizing all variants to "SLES 11 systems"

Bischoff:
Let's agree on SLES xx everywhere (capitals, space)


Bischoff:
dario is fixing, dropping pr

PR #451, opened 2017-09-22
Normalizing all variants to "SLES 11 systems"

Bischoff:
Let's agree to use SLES xxx everywhere


Bischoff:
dario is fixing, dropping pr

PR #454, opened 2017-10-05
[WIP] Manager31 minion not in configuration -> target systems list

hustodemon:
will also port to `master`


MalloZup:
@hustodemon  make rubocop happy first
```ruby
features/step_definitions/navigation_steps.rb:311:43: C: Prefer to_s over string interpolation.

  find('input[name="filter_string"]').set("#{$minion.full_hostname}")

                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^

features/step_definitions/navigation_steps.rb:312:35: C: Do not use parentheses for method calls with no arguments.

  find('button[value="Go"]').click()

                                  ^
```

features/step_definitions/navigation_steps.rb
```diff
@@ -302,6 +302,16 @@
   find_link('Configuration').click
 end
 
+Given(/^I am on the Configuration => Systems => Target systems page$/) do
+  step %(I am authorized)
+  visit("https://#{$server.full_hostname}/rhn/configuration/system/TargetSystems.do")
+end
+
+Given(/^I type "sle-minion" in the search box and click "Go"$/) do
```
MalloZup: we are hardcoding a parameter here. ( you need a general one step here) where we pass the regex and put them in a var
https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/step_definitions/navigation_steps.rb#L220

MalloZup:
here is the naming convention for secondary features:
https://github.com/SUSE/spacewalk-testsuite-base/blob/master/docs/Guidelines.md#rules-for-features

example
https://github.com/SUSE/spacewalk-testsuite-base/blob/master/run_sets/testsuite.yml#L54

features/step_definitions/navigation_steps.rb
MalloZup: additionaly, you could even make the 'GO" click a parameter.

hustodemon:
Ufff, sorry for the embarrassment with the `franky.feature`! :trollface:  Fixed your issues, but didn't add to `refhost` - is there a reason why this test should run on ref hosts?

guys, let me know if it's ok now. If yes, i'll squash the commits, merge it and backport to `master`.

features/min_not_in_target_systems.feature
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2017 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Systems in "Configuration => Systems => Target Systems" page
```
MalloZup: What is the feature goal here?

features/min_not_in_target_systems.feature
```diff
@@ -0,0 +1,9 @@
...
+
+  Scenario: Check absence of the minion in the list
```
MalloZup: Why the minion is absent and not present? 

I am just thinking if someday it is present, or if we switch the feature order, can work? why we expect that the minion is absent

MalloZup:
@hustodemon  thx ! the ruby code steps are fine.

I would however:
- add the feature to an existing secondary feature.
- rename the feature name, and explain what is the goal of the test. Why are we expecting to see not target system found? If we change the order of the feature, does still work?
- what are the dependency of this test?

In longterm all this seems you contradictory, but we need to make clear th intent of the test and which dependency are implicit, to don"t have a failing test without know why. Just looking in the logic there, we are expecting that there are no system in that page, but ok, what does it mean?

Bischoff:
"Explain the goal of the test" : yes, but please do this in the scenario name - NOT in the feature description

The feature description is for explaining which feature OF THE PRODUCT we test, not to describe the test.

I felt I needed to clarify this @MalloZup :smile:

MalloZup:
sure , buti it not clear which feature are we testing even. ( a part of that)

Bischoff:
Right.

hustodemon:
> add the feature to an existing secondary feature.

Can you advise to which one? I was looking in the features, but couldn't find any place where this would fit, that's why I created a new one.

> rename the feature name

I was looking at a couple of other `Feature`s and i thought the name "Systems in "Configuration => Systems => Target Systems" page" was quite ok comparing to them (e.g. "Patches page", "The channels page"). Then OTOH, we have features that are called like "Install a patch on the client via Salt through the UI", or "Install a package to the traditional client", but this only looks like we don't have any convention here at all! What should be then the right name for a feature? Should it be a sentence, or a name?

> and explain what is the goal of the test. Why are we expecting to see not target system found?

Salt minions (unlike traditional systems) should never appear in that list. I'll try to be more explicit in the `Scenario`.

> If we change the order of the feature, does still work?  what are the dependency of this test?

It depends on the presence of a salt minion in SUMA. I assumed that the core features ensure that. Don't they?

MalloZup:
@hustodemon thank you for the honest feedback, yes. :clap: :clap: 

Ok , now we know that  a salt-minion should never appear there. 
1. So the feature we need to adress is here:
  https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/allcli_overview_systems_details.feature

 I would personally add this kind of test for all 3 types of Minions , creating for each a scenario.

2. For the Feature Naming, it was always a mess the naming of Scenarios. for the scenario Names, there are small conventions https://github.com/SUSE/spacewalk-testsuite-base/blob/master/docs/Guidelines.md#rules-for-scenarios.

I completely agree with you, but we are starting to change it, and we want new test to be less messy that in the past.
 For me there, the most important it to understand what the test want to do.
 This is the most important stuff for maintaining in long term.  
I am not so pedant for that

3. > It depends on the presence of a salt minion in SUMA. I assumed that the core features ensure that. Don't they? 

  Correct! :rocket:

MalloZup:
@hustodemon i am closing this pull request because this repo will move inside the spacewalk fs repo.

Feel free to reopen it against SUSE/spacewalk soon. thx

PR #459, opened 2017-10-18
[WIP] Proxy support - master

Bischoff:
**Do not merge**

This PR is just to show work in progress :airplane:

### DIFF
--- a/docs/static-run.md
+++ b/docs/static-run.md
@@ -10,6 +10,7 @@ With a static setup, if you break your testing machines, or if you have any trou
 Set up the following environment variables:
 
 * TESTHOST the SUSE Manager server you are testing against
+* PROXY the SUSE Manager proxy (use "None" if there is no proxy)
 * CLIENT the traditional client
 * MINION the Salt minion
 * SSHMINION the SSH-managed Salt minion
@@ -20,6 +21,7 @@ To run all standard tests, from the controller:
 
 ```console
 export TESTHOST="${PREFIX}suma3pg.tf.local"
+export PROXY="None"
 export CLIENT="${PREFIX}clisles12sp3.tf.local"
 export MINION="${PREFIX}minsles12sp3.tf.local"
 export SSHMINION="${PREFIX}minsles12sp3ssh.tf.local"
--- a/features/step_definitions/common_steps.rb
+++ b/features/step_definitions/common_steps.rb
@@ -483,7 +483,12 @@ Given(/^I update the profile of this client$/) do
 end
 
 When(/^I register using "([^"]*)" key$/) do |arg1|
-  regurl = "http://#{$server_ip}/XMLRPC"
+  # register to the proxy, or if none, to the server
+  regurl = if $proxy.nil?
+    "http://#{$server_ip}/XMLRPC"
+  else
+    "http://#{$proxy_ip}/XMLRPC"
+  end
   command = "rhnreg_ks --force --serverUrl=#{regurl} --activationkey=#{arg1}"
   $client.run(command, true, 500, 'root')
 end
--- a/features/step_definitions/navigation_steps.rb
+++ b/features/step_definitions/navigation_steps.rb
@@ -406,6 +406,22 @@ Then(/^I should see something$/) do
     )
 end
 
+Then(/^I should see no system apart from the proxy$/) do
+  if $proxy.nil?
+    step %(I should see a "No systems." text)
+  else
+    step %(I should see a "1 - 1 of 1" text)
+  end
+end
+
+Then(/^I should see no virtual system apart from the proxy$/) do
+  if $proxy.nil?
+    step %(I should see a "No Virtual Systems." text)
+  else
+    step %(I should see a "1 - 2 of 2" text)
+  end
+end
+
 Then(/^I should see "([^"]*)" systems selected for SSM$/) do |arg|
   within(:xpath, '//span[@id="spacewalk-set-system_list-counter"]') do
     raise unless has_content?(arg)
--- a/features/support/twopence_init.rb
+++ b/features/support/twopence_init.rb
@@ -1,13 +1,16 @@
 require 'twopence'
 
 # Initialize SSH targets from environment variables
+# If there is no proxy, use "None" for $PROXY environment variable
 raise 'Server IP address or domain name variable empty' if ENV['TESTHOST'].nil?
+raise 'Proxy IP address or domain name variable empty' if ENV['PROXY'].nil?
 raise 'Client IP address or domain name variable empty' if ENV['CLIENT'].nil?
 raise 'Minion IP address or domain name variable empty' if ENV['MINION'].nil?
 raise 'CentOS minion IP address or domain name variable empty' if ENV['CENTOSMINION'].nil?
 raise 'SSH minion IP address or domain name variable empty' if ENV['SSHMINION'].nil?
 
 $server_ip = ENV['TESTHOST']
+$proxy_ip = ENV['PROXY']
 $client_ip = ENV['CLIENT']
 $minion_ip = ENV['MINION']
 $ceos_minion_ip = ENV['CENTOSMINION']
@@ -15,6 +18,7 @@ $ssh_minion_ip = ENV['SSHMINION']
 
 # Define twopence objects
 $client = Twopence.init("ssh:#{$client_ip}")
+$proxy = Twopence.init("ssh:#{$proxy_ip}") if $proxy_ip != 'None'
 $server = Twopence.init("ssh:#{$server_ip}")
 $minion = Twopence.init("ssh:#{$minion_ip}")
 $ceos_minion = Twopence.init("ssh:#{$ceos_minion_ip}")
@@ -23,14 +27,16 @@ $ssh_minion = Twopence.init("ssh:#{$ssh_minion_ip}")
 # Lavanda library module extension
 # Look at support/lavanda.rb for more details
 $server.extend(LavandaBasic)
+$proxy.extend(LavandaBasic) if $proxy
 $client.extend(LavandaBasic)
 $minion.extend(LavandaBasic)
 $ceos_minion.extend(LavandaBasic)
 $ssh_minion.extend(LavandaBasic)
 
 # Initialize hostname
-nodes = [$server, $client, $minion, $ceos_minion, $ssh_minion]
+nodes = [$server, $proxy, $client, $minion, $ceos_minion, $ssh_minion]
 nodes.each do |node|
+  next if node.nil?
   hostname, _local, _remote, code = node.test_and_store_results_together('hostname', 'root', 500)
   raise 'Cannot get hostname for node' if code.nonzero?
   node.init_hostname(hostname)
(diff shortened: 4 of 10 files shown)


MalloZup:
:rocket:

docs/static-run.md
```diff
@@ -20,6 +21,7 @@ To run all standard tests, from the controller:
 
 ```console
 export TESTHOST="${PREFIX}suma3pg.tf.local"
+export PROXY="None"
```
MalloZup: can we use 'NOT_SET' or something else then 'None' because in python and ruby lead to misbehaviour with the nil object. :rocket:

features/support/twopence_init.rb
```diff
@@ -1,50 +1,57 @@
...
+# If there is no proxy, use "None" for $PROXY environment variable
+raise 'Server IP address or domain name variable empty' if ENV['TESTHOST'].nil?
+raise 'Proxy IP address or domain name variable empty' if ENV['PROXY'].nil?
+raise 'Client IP address or domain name variable empty' if ENV['CLIENT'].nil?
+raise 'Minion IP address or domain name variable empty' if ENV['MINION'].nil?
+raise 'CentOS minion IP address or domain name variable empty' if ENV['CENTOSMINION'].nil?
+raise 'SSH minion IP address or domain name variable empty' if ENV['SSHMINION'].nil?
 
 $server_ip = ENV['TESTHOST']
+$proxy_ip = ENV['PROXY']
```
MalloZup: it would be great if we get rid of Global variables, but ok  for time beeing

docs/static-run.md
MalloZup: I would prefer over all a boolean var false

features/support/twopence_init.rb
```diff
@@ -1,50 +1,57 @@
...
 $client_ip = ENV['CLIENT']
 $minion_ip = ENV['MINION']
 $ceos_minion_ip = ENV['CENTOSMINION']
 $ssh_minion_ip = ENV['SSHMINION']
 
-# define twopence object.
+# Define twopence objects
 $client = Twopence.init("ssh:#{$client_ip}")
+$proxy = Twopence.init("ssh:#{$proxy_ip}") if $proxy_ip != 'None'
```
MalloZup:
``` unless proxy_ip

File: features/support/twopence_init.rb (hunk 2 above)
Comment (MalloZup) [reply features/support/twopence_init.rb]: this one can be not global

File: features/support/twopence_init.rb [hunk 4]
```diff
@@ -1,50 +1,57 @@
...
+$proxy.extend(LavandaBasic) if $proxy
 $client.extend(LavandaBasic)
 $minion.extend(LavandaBasic)
 $ceos_minion.extend(LavandaBasic)
 $ssh_minion.extend(LavandaBasic)
 
-# add here new vms .
+# Initialize hostname
 nodes = [$server, $client, $minion, $ceos_minion, $ssh_minion]
+nodes.push($proxy) if $proxy
```
MalloZup: i think something like this is more consistent, and more generic.
```console
nodes = [server, client, minion, ceos_minion, ssh_minion, proxy]
nodes.each do |node|
      # if node false just skip
      next unless node
end
```

MalloZup: i think something like this is more consistent, and more generic.
```console
nodes = [server, client, minion, ceos_minion, ssh_minion, proxy]
nodes.each do |node|
      # if node false just skip
      next unless node
end
```

run_sets/testsuite.yml
```diff
@@ -16,6 +16,7 @@
 - features/core_srv_users.feature
 - features/core_srv_credentials.feature
 - features/core_srv_create_activationkey.feature
+- features/core_proxy.feature
```
MalloZup: maybe we could name this
```  features/core_proxy_if_enabled.feature```
 or add a comment above to say that this doesnt run always

File: features/support/twopence_init.rb (hunk 2 above)
Comment (Bischoff) [reply features/support/twopence_init.rb]: No, it's used at other places

As you say, we should remove them, but thzen all of them

File: docs/static-run.md (hunk 1 above)
Comment (Bischoff) [reply docs/static-run.md]: "Non" is what Salt answers if a grain is not set. That's why I am using this.

We have a kind of boolean: $proxy.nil?

File: features/support/twopence_init.rb (hunk 3 above)
Comment (Bischoff) [reply features/support/twopence_init.rb]: no, it is "None" - dictated by Salt

File: features/support/twopence_init.rb (hunk 4 above)
Comment (Bischoff) [reply features/support/twopence_init.rb]: ok

File: run_sets/testsuite.yml (hunk 5 above)
Comment (Bischoff) [reply run_sets/testsuite.yml]: this comment is inside of the file

but the name could indeed be better

Speaking of which,

     core_setup_user_and_proxy.feature

has now become misleading (it's about HTTP proxy, not suma proxy)

File: features/support/twopence_init.rb (hunk 3 above)
Comment (MalloZup) [reply features/support/twopence_init.rb]: @Bischoff can you point me where in salt you get this Grains variable?

File: features/support/twopence_init.rb (hunk 4 above)
Comment (Bischoff) [reply features/support/twopence_init.rb]: Done.

File: features/step_definitions/common_steps.rb [hunk 6]
```diff
@@ -483,7 +483,12 @@ def enable_pool(target)
 end
 
 When(/^I register using "([^"]*)" key$/) do |arg1|
-  regurl = "http://#{$server_ip}/XMLRPC"
+  # register to the proxy, or if none, to the server
+  if $proxy_ip == 'None'
+    regurl = "http://#{$server_ip}/XMLRPC"
+  else
+    regurl = "http://#{$proxy_ip}/XMLRPC"
```
MalloZup: dont use $proxy_ip here.
we should introduce $proxy.ip .

If you add a method in this module, we can have it
https://github.com/SUSE/spacewalk-testsuite-base/blob/master/features/support/lavanda.rb#L7

MalloZup: with this we can remove the global variable

features/support/twopence_init.rb
MalloZup: i would rather create a method for ip , so we don introduce new unclean code..

features/step_definitions/common_steps.rb
Bischoff: I suggest we keep refactoring work like this separate from this PR.

run_sets/testsuite.yml
MalloZup: Remove the name, for that feature. The stuff inside the file, we need something outside, for the global picture when reading all feature tog

MalloZup: Remove the name, for that feature. The stuff inside the file, we need something outside, for the global picture when reading all feature tog

features/support/twopence_init.rb
Bischoff: sumaform/salt/controller/bashrc
```
export PROXY={{ grains.get("proxy") }}
```

"None" is not harcoded here - it's what Salt gives when the grain is missing.

features/support/twopence_init.rb
Bischoff: Yes, but let's do this kind of refactorings in separate PRs. This is complicated and big enough.

run_sets/testsuite.yml
Bischoff: "Remove the name, for that feature." Which feature, the HTTPS proxy or the SUMA proxy one?

And remove how?

features/support/twopence_init.rb
MalloZup: There is a way to  set defaults in saltstack when the grains is None. 
atm cannot remember

run_sets/testsuite.yml
```diff
@@ -16,76 +16,5 @@
...
-- features/min_salt_software_states.feature
-- features/trad_metadata_check.feature
-- features/srv_clone_channel_npn.feature
-- features/trad_cve_id_new_syntax.feature
-- features/trad_weak_deps.feature
-- features/srv_nagios.feature
-- features/srv_cve_audit.feature
-- features/min_salt_install_with_staging.feature
-- features/srv_xmlrpc_activationkey.feature
-- features/allcli_overview_systems_details.feature
```
MalloZup: remember to dont  eliminate  this

features/step_definitions/navigation_steps.rb
```diff
@@ -406,6 +406,14 @@
     )
 end
 
+Then(/^I should see no system apart from the proxy$/) do
+  if $proxy.nil?
```
MalloZup: use guard clause

run_sets/testsuite.yml
Bischoff: ??? why would i do that ?

features/step_definitions/navigation_steps.rb
MalloZup: http://www.thechrisoshow.com/2009/02/16/using-guard-clauses-in-your-ruby-code/

run_sets/testsuite.yml
MalloZup: in you pr you are removing this features

run_sets/testsuite.yml
MalloZup: ```core_setup_user``` is fine

features/step_definitions/navigation_steps.rb
Bischoff: Thanks. Fixed this one with a ternary, looks more adapted to me.

features/step_definitions/common_steps.rb
```diff
@@ -483,7 +483,12 @@ def enable_pool(target)
 end
 
 When(/^I register using "([^"]*)" key$/) do |arg1|
-  regurl = "http://#{$server_ip}/XMLRPC"
+  # register to the proxy, or if none, to the server
```
MalloZup: http://www.thechrisoshow.com/2009/02/16/using-guard-clauses-in-your-ruby-code/

run_sets/testsuite.yml
Bischoff: Right, deserves a separate PR. I have no idea of correct name but i'll find one.

run_sets/testsuite.yml
Bischoff: "oops"

:blush:

features/step_definitions/common_steps.rb
Bischoff: I did not touch that one yet. But it might come, for sure.

Bischoff: replaced with a ternary

features/step_definitions/navigation_steps.rb
Bischoff: honestly, for that one, a guard clause is less readable and does not bring anything

run_sets/testsuite.yml
Bischoff: done, with `core_first_settings` (it's not only about users, but also organization, http proxy, etc)

Bischoff:
Closing, because I can't even read anymore my own code, too many discussions ;-)

MalloZup:
@Bischoff  you have the button hide comments

PR #462, opened 2017-10-19
Support for proxy - master

Bischoff:
This pull request adds support for proxy in the test suite.

It is incomplete, i.e. it stops at using the proxy for the trad client and the salt minion. The other bootstraps have not been modified yet. This will be done in some other PR.

The test suite behaviour is left unchanged if shell variable PROXY is not set.


features/support/commonlib.rb
```diff
@@ -34,3 +34,31 @@ def filename
...
+
+# generate answers file to interactively run configure-proxy.sh.
+def generate_answers_file(server, proxy)
+  raise(ArgumentError, 'Hostname is missing!') if server.nil? || server.empty? || proxy.nil? || proxy.empty?
+  path = ''
+  Tempfile.open('config-answers.txt') do |f|
+    path = f.path
+    f.write("RHN_PARENT=#{server}\n")
+    f.write("HTTP_PROXY=''\n")
+    f.write("VERSION=3.2\n")
```
MalloZup: Nitpick: remember to dont hardcode 3.2 Version number, since we want to chose servers

features/step_definitions/common_steps.rb
```diff
@@ -483,7 +483,12 @@ def enable_pool(target)
 end
 
 When(/^I register using "([^"]*)" key$/) do |arg1|
-  regurl = "http://#{$server_ip}/XMLRPC"
+  # register to the proxy, or if none, to the server
+  regurl = if $proxy.nil?
```
MalloZup: In ruby there  is also the ? operator, which you could use also. 
```regurl = if $proxy nil? ?   "http://#{$server_ip}/XMLRPC" : ` "http://#{$proxy_ip}/XMLRPC"``

File: features/support/commonlib.rb (hunk 1 above)
Comment (Bischoff) [reply features/support/commonlib.rb]: This is "head" branch, I was planning to change this to 3.1 on "manager31" branch.

So there's not a big problem to hardcode if we use a different version on each git branch.

File: features/step_definitions/common_steps.rb (hunk 2 above)
Comment (Bischoff) [reply features/step_definitions/common_steps.rb]: I tried ? :  initialy, but it looked a bit too obfuscated...

Comment (MalloZup) [reply features/step_definitions/common_steps.rb]: Ok

File: features/support/commonlib.rb (hunk 1 above)
Comment (MalloZup) [reply features/support/commonlib.rb]: The problem with this git branch isthat imagine that HEAD -> 3.3 . 
I think we should Pick up this variable from a file that susemanager offer or via spacecmd XMLRPC api if there is not a file that reveals the suse-manager version

Comment (MalloZup) [reply features/support/commonlib.rb]: /usr/share/rhn/config-defaults # vim rhn_web.conf 
or 
using rpm -qa susemanager

Comment (Bischoff) [reply features/support/commonlib.rb]: We can change this to 3.3 when head moves to 3.3, no big deal.

BTW in sumaform this value is still.... 2.1 !!! Apparently, this has no big consequences. I even wonder if it could be removed from the product.

Comment (MalloZup) [reply features/support/commonlib.rb]: Ok thx ebischoff for the investigation. I think it is worth to ask if we can remove it.
If we have the possibilty i would not add a manual change to versions, this is simply unsafe, imagine  we start to rely on all this small, changes

**Comment by MalloZup** [comment]
https://github.com/moio/sumaform/pull/202/commits/f9c4d20d92610489acafd8af243b01ee6aa86181#diff-0bf3346ebbff8e21e0e418df485ea0c6R20

File: docs/static-run.md [hunk 3]
```diff
@@ -20,6 +21,7 @@ To run all standard tests, from the controller:
 
 ```console
 export TESTHOST="${PREFIX}suma3pg.tf.local"
+export PROXY="None"
```
MalloZup: https://github.com/moio/sumaform/pull/202/commits/f9c4d20d92610489acafd8af243b01ee6aa86181#diff-0bf3346ebbff8e21e0e418df485ea0c6R20

Bischoff: it's what salt gives when the vm does not exist

features/step_definitions/command_steps.rb
```diff
@@ -358,6 +361,46 @@
...
+    rc = file_inject($proxy, '/tmp/' + file, '/root/ssl-build/' + file)
+    raise 'File injection failed' unless rc.zero?
+  end
+end
+
+When(/^I configure the proxy$/) do
+  # prepare the settings file
+  settings = "RHN_PARENT=#{$server_ip}\n" \
+             "HTTP_PROXY=''\n" \
+             "VERSION=3.2\n" \
```
MalloZup: we need to take the version var dinamically from a file, otherwise this is a risk.

features/step_definitions/command_steps.rb
```diff
@@ -358,6 +361,46 @@
...
+    raise 'File extraction failed' unless rc.zero?
+    $proxy.run('mkdir -p /root/ssl-build')
+    rc = file_inject($proxy, '/tmp/' + file, '/root/ssl-build/' + file)
+    raise 'File injection failed' unless rc.zero?
+  end
+end
+
+When(/^I configure the proxy$/) do
+  # prepare the settings file
+  settings = "RHN_PARENT=#{$server_ip}\n" \
```
MalloZup: settings should be more a ruby constant (uppercase)

```ruby
SETTINGS = 'blabla'.freeze
```

features/support/twopence_init.rb
```diff
@@ -41,22 +47,17 @@
...
-  when 'sle-minion'
-    node = $minion
-  when 'sle-client'
-    node = $client
-  when 'sle-migrated-minion'
-    node = $client
-  else
-    raise 'Invalid target.'
-  end
+  map = {
```
MalloZup: really  good refactoring! :+1:

MalloZup: 1. can we avoid to call map ( which is an implicit ruby built-in name) and we can use a more explicit name?

like ```node_hash``` instead of ``` map```

features/support/twopence_init.rb
```diff
@@ -41,22 +47,17 @@
...
+    'server' => $server,
+    'proxy' => $proxy,
+    'ceos-minion' => $ceos_minion,
+    'ssh-minion' => $ssh_minion,
+    'sle-minion' => $minion,
```
MalloZup: 2. the newer syntax in ruby is more readable

```ruby
 { key: 'value' }
'sle-minion': $minion,
```

features/support/twopence_init.rb
```diff
@@ -41,22 +47,17 @@
...
+    'sle-client' => $client,
+    'sle-migrated-minion' => $client
+  }
+  node = map[host]
```
MalloZup: renaming needed, but awesome refactoring

features/support/setup_browser.rb
```diff
@@ -7,7 +7,7 @@
 require 'base64'
 require 'capybara'
 require 'capybara/cucumber'
-require File.join(File.dirname(__FILE__), 'cobbler_test')
+require_relative 'cobbler_test'
```
MalloZup: +1

features/step_definitions/command_steps.rb
Bischoff: In ruby, uppercase is for constants. This is a variable.

features/step_definitions/command_steps.rb
Bischoff: We have a different branch for each version. The only risk is to forget to bump this version number.

But I'll do it to please you.

features/support/twopence_init.rb
Bischoff: OK.

features/support/setup_browser.rb
Bischoff: LOL for the description of the style. But please defuckify your comments :-).

I am not sure about the require_relative not being needed. I suggest that if you want to experiment with this, you do it in a separate pull request.

features/step_definitions/command_steps.rb
MalloZup: We can let this,  and if we forget to bump, you pay me 1000 euro,

features/step_definitions/command_steps.rb
MalloZup: This a constant variable. You dont change it later after init.

MalloZup: But ok

features/support/setup_browser.rb
MalloZup: Create a fixme commenr above the code so we can adress it

features/support/twopence_init.rb
Bischoff: Right.

features/step_definitions/command_steps.rb
Bischoff: Then I change it :fox_face:

features/support/setup_browser.rb
Bischoff: ok

features/step_definitions/command_steps.rb
MalloZup: :wolf: :+1:

Bischoff: 1) I could let sumaform (or the user) create another variable

export PROXY_VERSION=3.1

but it seems too much for something that is probably not even used (we've been using "2.1" here for years without any consequence).

2) Or maybe I could just set it to "", which is "same version as server".

I'll try the latest option.

Bischoff: ok it works with this

```
ebi-proxy:~ # cat config-answers.txt20171025-5889-yxg00i 
RHN_PARENT=ebi-server.tf.local
HTTP_PROXY=''
VERSION=''
(...)
```

features/support/twopence_init.rb
MalloZup: create this like that 

```ruby
options = { font_size: 10, font_family: "Arial" }
````

File: features/support/setup_browser.rb (hunk 9 above)
Comment (MalloZup) [reply features/support/setup_browser.rb]: @Bischoff  where is the fixme here?

File: features/support/twopence_init.rb (hunk 7 above)
Comment (Bischoff) [reply features/support/twopence_init.rb]: Not sure I understand.
```
sle-minion: $minion
```
? Please notice "sle-minion" is a string, not a key.

Comment (MalloZup) [reply features/support/twopence_init.rb]: why do you need a string: 

```ruby

#! /usr/bin/ruby

def get_target(host)
  $server = 'baubau'
  $proxy = 'mia'
  nodes_map = {
    server:  $server,
    proxy:  $proxy
  }
  node = nodes_map[host.to_sym]
  puts node
  node
end

grande_host = 'server'

get_target(grande_host)
```

features/support/setup_browser.rb
Bischoff: just below.

features/support/twopence_init.rb
Bischoff: No. `$server` contains a node, not a string. I cannot do $server = 'baubau'.

Sorry, i stop losing time with this -- `=>` is just fine.

Please come up with a working and tested solution and prepare a pull request if youfind the way to do it. But I won't spend more time on something's that just cosmetic.

Bischoff:
Bug with the extraction of `/tmp/spacewalk-debug.tar.bz2` is now fixed. We can merge.

features/support/twopence_init.rb
MalloZup: the solution i provided it works. For ruby having a Twopence Node or Baubau Object doesn't matter.
(Baubau was an example sorry!! ) 
At this point i will refactor it later, if you cannot and are fine

Bischoff: thanks for your undestanding
:+1:

MalloZup:
:rocket:

PR #473, opened 2017-10-25
Support for proxy - manager31

Bischoff:
This pull request adds support for proxy in the test suite.

It is incomplete, i.e. it stops at using the proxy for the trad client and the salt minion. The other bootstraps have not been modified yet. This will be done in some other PR.

The test suite behaviour is left unchanged if shell variable PROXY is not set.


Bischoff:
Bug with the extraction of /tmp/spacewalk-debug.tar.bz2 is now fixed. We can merge.

PR #472, opened 2017-10-25
Manager-3.1: Verify that "Admin" is visible only for admin users (bsc#1063822)

renner:
This patch is supposed to verify that the "Admin" link in the main navigation is visible for administrator users only. The "testing" user should not see the "Admin" link.

PR #474, opened 2017-10-25
Master: Verify that "Admin" is visible only for admins (bsc#1063822)

renner:
This is a forward-port of #472 for the master branch. Probably no additional review required as there is no difference to the `manager31` version of the patch.


MalloZup:
@renner ok

